PR #177, opened 2014-02-10
Added Catalan translation

cescp:
Please find in this request the addition of Catalan translation (still in progress) for NOOBS.


XECDesign:
See https://github.com/raspberrypi/noobs/pull/373
Thank you

PR #199, opened 2014-06-29
Add for Korean Translation

sople1:
Hi, i work finished just few minutes, and i want to push .ts file for this project.
ah, when i run pre-commit-translation-update-hook.sh , this script edits another ts files, so i reverted and forked to only add korean ts file. is it correct?

add: can i compile on OSX environment? i want to test this ts file but i don't know how can i test on pi environment.


XECDesign:
Fixed by d80ed158f6463c3a91c2f7069481360f1652904c, thank you.

Issue #213, opened 2014-09-15
NOOBS installation size idea - modularize!


procount:
The distros used with NOOBS are getting so large that it is difficult to fit them all onto a 4GB SD card.
NOOBS v1.3.10 has therefore removed all but Raspbian, with the remainder having to be downloaded "on the fly", so it is now closer in function to NOOBS-lite.
Here is an alternative suggestion to harmonise the installation process so that it is suitable for different sizes of SD card or installation preferences.
1. Make NOOBS-lite the primary release of NOOBS. i.e. just the installer with no included OSs. User can unzip this file to a blank FAT32 SD card as now. User could use this as is and download/install any OS from within NOOBS as now.
2. Make each OS to be installed by NOOBS available on the website in a NOOBS installable format. i.e. as .tar.xz files, complete with all the installation slides, .json files etc. and package them up into a zip file for each distro.
3. User can download the zip file(s) for whichever OS distros they want to install and unzip them over the top of their NOOBS-lite installer on the SD card. User could now use this SD card to install those OS to a PI when not connected to a network. If connected to the internet, NOOBS lite could additionally offer the other OS available for download as now.
4. Modify NOOBS to look for OSs on /dev/sda1 in addition to /dev/mmcblk0p1 and the internet. In this way a user can have a minimal NOOBS image on their SD card, with more available space on it for their Distros. They can keep all their OS installation files on a separate USB flash drive, to be plugged in just when they want to re-install. In this way they can still do a recovery option when away from the Internet.

Of course, steps 1-3 could be done by an average user now, but packaging the distros as a zip file and making them available on the website for download would make it easier for newbies to select and download. 
Step 4 needs a modification to NOOBS, but I think it would bring a lot of flexibility without overcomplicating the installation.

Ruffio:
@procount is this still an issue?

procount:
Partially solved by @robbi5, but it would be nice to allow users to download a zip or tar file for each distro so they can choose which to download and integrate into Noobs-lite.

procount:
A similar download feature has been added to PINN.

Issue #164, opened 2013-12-30
[feature request] Installing OS from external storage device


gyeben:
I think it would be great if operating systems could be installed from an external storage device, eg: a pendrive or an USB HDD.

PR #219, opened 2014-11-17
Add Installation of OS Images from USB

robbi5:
Hi,

for various use cases like
- users with a small SD card
- sharing one usb pendrive in a classroom
- (re)installing on an raspberry pi compute module
- ...?

it would be great to use os images stored on an USB storage device.
This PR mounts the first partition of an FAT formatted USB drive (`/dev/sda1`) if available and looks for additional os images in the `os` folder on it.

I've seen other people looking for something like this too: #164 and #213


robbi5:
Hi,
thanks for the buildroot update :+1: 
A fresh version of the changes on top of NOOBS v1.3.12 is in [headissue:usb-os-install-pi2](https://github.com/headissue/noobs/tree/usb-os-install-pi2).
Should we close this PR and open a new one?

lurch:
It's up to you - you can either close this PR and open a new one; or you can do an interactive git rebase to push your changes into this PR (your headissue:usb-os-install branch).

I can't make any promises yet when / if this PR will get merged into mainline though.

ghollingworth:
Might try merging this into Slice NOOBS since it would be useful to be
able to update from a USB stick

Gordon

procount:
Is it necessary to include the line "# CONFIG_WLAN is not set" in both the kernel config files? I think removing this line from each will not affect the build, but will allow easier integration with #243 - the wifi addition. Please correct me if I am wrong.

lurch:
IIRC if you start manually taking out the `# XXX is not set`, then when you build the kernel it assumes that this is a new not-seen-before setting, and then asks you whether it should be enabled or not. Which can then cause problems with the buildroot process, as it doesn't expect the build process to require interactive intervention.

The contents of the kernel config files are determined automatically by the kernel's `defconfig` Makefile target (see #243 for a discussion of how this has been integrated into the dual-kernel setup of NOOBS).

procount:
I originally thought that "# CONFIG_WLAN is not set" was just a comment, but it seems it actively disable the wifi configuration. When integrated with #243 all the wifi functionaltiy stopped working and removing this line resolves the issue. Interestingly, there is no CONFIG_WLAN setting in the kermel config files for #243, which I guess is why there was no merge conflict to flag up this issue. Perhaps the issue is that #243 should have "CONFIG_WLAN=y" ???

procount:
Probably this is just my integration issue. Does "# CONFIG_WLAN is not set" get removed and replaced by its sub-menu options when it is selected? In which case I need to take care when merging the kernel-recovery files that you get one option or the other but not both, which a simple merge will not identify.

lurch:
This is all very much AIUI (I might be wrong), but I believe the 'defconfig' is automatically created as the 'minimal set of config options' needed for a specific kernel configuration. As such, `CONFIG_RT2X00=y` in e.g. `buildroot/kernelconfig-recovery.armv6` *implies* `CONFIG_WLAN=y` and so there's no need to explicitly include the CONFIG_WLAN option. (I believe when you 'apply' the defconfig, it automatically creates the full every-option-included .config file)

> When integrated with #243 all the wifi functionaltiy stopped working and removing this line resolves the issue.

I guess the lesson here is that you can't simply merge 'defconfig' files...? Perhaps there is a way to do what you want 'automatically' but I'm afraid I dunno what that is.

Pinging @popcornmix as he seems to be very familiar with manipulating the defconfig files for https://github.com/raspberrypi/linux

procount:
Maybe it is necessary to create the full .config file for each defconfig option, merge and resolve the full .config files, then create a defconfig file from that?

lurch:
> Maybe it is necessary to create the full .config file for each defconfig option, merge and resolve the full .config files, then create a defconfig file from that?

Sounds plausible :)

procount:
This is now supported in NOOBS 2.4, so I think this can be closed.

XECDesign:
Thanks

Issue #231, opened 2015-02-10
option to override hardcoded DEFAULT_REPO_SERVER


beta-tester:
it would be nice, to be able to override the hardcoded
`#define DEFAULT_REPO_SERVER "http://downloads.raspberrypi.org/os_list_v2.json"`
for special purpose.
for example by adding an option with the link to the local server in the cmdline.txt file or any other file responsable for noobs.

one of the special purpose is to download all the OS installation files once from RaspberryPi webpage to a local server and add new or modify existing images to share/distribute them over the local network.
so all local RPi's can be installed quick and easily with the modified images in an education environment.

or sometimes the schools network firewall will not let the RPi to "go outside"...
so a local solution would be nice...
i know, preparing a "heavy loaded" NOOBS version SD card with all needed images on it would be a solution, but it is waste of SD card space and installation preparation time for the teachers...
a quick installation of a NOOBS lite and installation via local network would be possible for the childs in-time of a lession easily

procount:
I can do that for you. In fact I've already done it in my version, but it's based on 1.3.10, not the latest 1.3.12 for the Pi2. If that's suitable for you, I'll extract it from my extra mods and post a version up for you. Please allow a few days..... Would you prefer the updated source, or a pre-built version of noobs-lite?

beta-tester:
thank you...
if you have added that feature and done changes on the source-code already, then it is possibly a good idea, to make a kind of commit / pull request here, so your changes can go to the NOOBS master trunk...
so every body will benifit form your work... not only me.

procount:
Absolutely. But that may take a while because I think Lurch is a bit busy at the moment and it might have to align with RPF releases? In any case I will push it to my github fork, so anyone can build it from there and maybe Lurch can pull the changes through too. I'll post here when it's done. Just let me know if you need it prebuilt.

lurch:
I've also had this feature privately requested too... but in that case they wanted to be able to add their own server 'on top' of `DEFAULT_REPO_SERVER`, rather than replace it entirely. Any chance you could incorporate that option into your version too @procount, while you're at it? ;-)

procount:
I foresee a slippery slope ahead..... :-)

procount:
Please use Pull Request in #234 instead of #232. The latter could be deleted.

lurch:
@procount You can do that yourself - simply click on the "Close pull request" button on #232  :-)

beta-tester:
@procount and @lurch thank you alot...
[...]

PR #232, opened 2015-02-13
Issue 231

procount:
Adds 2 cmdline options: default_repos_server allows to change the source of OS images on the internet
and extra_repos_server allows an additional source of OS images to be used aswell. Solves issue #231


ghollingworth:
@procount I think it's my fault but your pull request has been a bit broken due to me accidentally deleting the head yesterday!  I put the commits back in again, but it seems it is making your pull request try to also merge the changes @lurch made for Pi 2.

Can you try merging the latest into your branch and squashing the resultant commits to a single one...  It looks fairly easy (and very useful since I'm developing Slice NOOBS and have basically just needed to add SSH and external json support myself!)

Gordon

procount:
@ghollingworth Oops!. OK, I was already in discussion with @lurch about modifying the approach and tidying up the commits, so I have created another PULL REQUEST to replace this one called alt_image_source #234. Please see if that does what you want. It is also based off the v1.3.12 pi2 version, but if necessary I can create another pull request off the older 1.3.10 version if that is more appropriate for the slice, although I thought the 1.3.12 version was suitable for Pi1 and Pi2?

procount:
Use Pull request #234 instead.

PR #234, opened 2015-02-17
Boolean selection of alternative image source repos. Replaces pull request: Issue_231

procount:
I changed the names of the parameters and it uses a boolean toggle as discussed with @lurch


procount:
Yep. No probs.

procount:
This modification provides an alternative URL to be specified to download the OS images from. Simply specify the URL of the additional os_list_v2.json file as the argument to the new alt_image_source parameter in recovery.cmdline. e.g. add <b>alt_image_source=http://newurl.com/os_list_v2.json </b> to the end of recovery.cmdline. This source of images will be appended to the default source. If you only want to use this alternative URL and suppress the default one, add <b>no_default_source</b> to recovery.cmdline.

lurch:
Just to clarify....
A while ago there was a change to RISC OS which resulted in only older versions of NOOBS being compatible with older versions of RISC OS, and only newer versions of NOOBS being compatible with newer versions of RISC OS.
I didn't want to break backwards compatibility, and older versions of NOOBS were already using `.../os_list.json`, so for newer versions of NOOBS I [switched](https://github.com/raspberrypi/noobs/commit/c779efa5801747d35006d8e51e41f6af78869c17) to using `.../os_list_v2.json`
I.e. old NOOBS -> `http://.../os_list.json` -> old RISC OS
and new NOOBS -> `http://.../os_list_v2.json` -> new RISC OS

**TL;DR** For the alt_image_source you'd be free to use any filename you want, including `alt_image_source=http://newurl.com/procounts_super_list_of_oses.json` ;-)
(or just `alt_image_source=http://newurl.com/os_list.json`)

procount:
Closed in order to fixup the commits. Will reopen shortly

PR #233, opened 2015-02-13
New ssh

procount:
Adds SSH support to noobs. Can be useful for debugging, troubleshooting, or headless installation. Need to add ssh to the cmdline to enable it. Wait a few seconds for the keys to be generated before you can connect.

PR #240, opened 2015-02-24
Added SSH

procount:
Adds SSH support to noobs. Can be useful for debugging, troubleshooting, or headless installation. Need to add ssh to the cmdline to enable it. Wait a few seconds for the keys to be generated before you can connect.
{replaces previous #233 pull request}

### DIFF
--- a/buildroot/.config
+++ b/buildroot/.config
@@ -1,6 +1,6 @@
 #
 # Automatically generated file; DO NOT EDIT.
-# Buildroot 2015.02-git-00009-g86ba8d2-dirty Configuration
+# Buildroot 2015.02-git-00001-gfff74f9-dirty Configuration
 #
 BR2_HAVE_DOT_CONFIG=y
 
@@ -1483,6 +1483,7 @@ BR2_PACKAGE_LIBICONV=y
 #
 # atftp needs a toolchain w/ threads, IPv6
 #
+# BR2_PACKAGE_AUTOSSH is not set
 # BR2_PACKAGE_AVAHI is not set
 # BR2_PACKAGE_AXEL is not set
 # BR2_PACKAGE_BANDWIDTHD is not set
@@ -1518,7 +1519,11 @@ BR2_PACKAGE_LIBICONV=y
 # BR2_PACKAGE_DHCPCD is not set
 # BR2_PACKAGE_DHCPDUMP is not set
 # BR2_PACKAGE_DNSMASQ is not set
-# BR2_PACKAGE_DROPBEAR is not set
+BR2_PACKAGE_DROPBEAR=y
+# BR2_PACKAGE_DROPBEAR_DISABLE_REVERSEDNS is not set
+BR2_PACKAGE_DROPBEAR_SMALL=y
+# BR2_PACKAGE_DROPBEAR_WTMP is not set
+# BR2_PACKAGE_DROPBEAR_LASTLOG is not set
 
 #
 # ebtables needs a toolchain w/ IPv6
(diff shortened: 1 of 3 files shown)


lurch:
Can you squash the commits? And then let me know the size of `recovery.rfs` both with and without this PR applied?

P.S. I tracked down https://github.com/raspberrypi/noobs/commit/acc5ec4d79de20930d1a93eda9954b108abe08f3#diff-540b0141b731233160e29daa56ad4d7dL1057 and https://github.com/raspberrypi/noobs/commit/acc5ec4d79de20930d1a93eda9954b108abe08f3#diff-f4eaeee2ea28031c3607d0f8842037c6L31 - does anyone ( @maxnet @ghollingworth ) remember why optional SSH support was removed originally?

NicoHood:
What about ssh support now?

procount:
Already done in my fork of NOOBS called PINN.
So feel free to pull it in, or just use PINN with its extra features.
See https://github.com/procount/pinn/blob/master/newfeatures.md

But note that this is for version 1.4 of NOOBS,
I have a local version for NOOBS v1.5 but I've not pushed it up to github yet.

NicoHood:
Thanks!
Why dont we add those improvements to noobs?

I thought about adding Ubuntu Mate and Arch (for Pi2) to noobs and customize a bit more. But there seems to be no blog post about hacking into noobs. And I never did something like that exactly before, so I have no real plan where to start. And compiling itself also fails #300 . Any idea where to get more infos on how to port such OS to noobs?

Also, YES I DO want to test your arch package. If you can also port ubuntu mate, it would be amazing! Just because all 3 OS have some pro and cons and I only have one very fast SD card at the moment.

NicoHood:
PINN does not contain any source files it seems, just the binaries.
I want to add SSH to my current noobs installation. that'd be nice.

procount:
@NicoHood - Yes, PINN has just the binaries. But all the source is in my NOOBS repo. The version of SSH for v1.5 has now been pushed too. I don't think there is anything unusual about Ubuntu Mate. You just need to repackage it as a custom OS as detailed in the NOOBS Readme.md.

Re: Hacking into NOOBS - I just spent ages reading and understanding the code before building it, and then modifying and testing it. Many people here have been very kind to help me along the way.

I put all of my mods on separate branches so that IF someone would like to include them in NOOBS, then it is easy to import them. But you have to convince the maintainers of their need and that they align with the general directives of NOOBS. Sometimes they are not just technical reasons... ;-)

NicoHood:
Okay, because from what I understood (for adding new images) is, that I then can only boot this single image. But I'd rather add ubuntu and arch to noobs instead.

procount:
You can use NOOBS to install many images onto your SD card and then boot any one of them. You are not restricted to just one. 
There is only one supplied with NOOBS full because they were just getting too big. The rest you need to install from the Internet. But you could put compressed versions of Raspbian, Arch and Ubunutu on your version of NOOBS and then install them all (if they all still fit!)

lurch:
Have a look at e.g. http://downloads.raspberrypi.org/NOOBS/images/NOOBS-2014-07-08/ which is the last version of "full NOOBS" that included more than just Raspbian, to see how multiple OSes can be supported by NOOBS when you don't have an internet connection.

NicoHood:
Why does the offline noobs installation has a boot/ copy inside the root/ tarball (of raspbian)?

lurch:
> Why does the offline noobs installation has a boot/ copy inside the root/ tarball (of raspbian)?

Hmmm, that's probably a bug (NOOBS v1.5.0 wasn't put together by me). I'll pass on the info.

RaspberryPiFan:
Why is this PR still open? Will this stay at PINN or be merged to NOOBS?

PR #243, opened 2015-03-01
Add wifi support - kernel supports Ralink rt2800usb drivers

jumaka:
Added early support for wifi into 1.4 noobs.
Added kernel driver support for the rt2800 devices as this is the network cards that I have access to.
Added a wifi access point selection screen and tested with WPA(2) networks - should work with open and WEP networks.
Will favour wired connections over wifi.
Have changed the wired code to make both wired and wireless consistent.
Tested on PI2 and original Model B.
Could not figure out how to trim the contents of kernelconfig_recovery... without getting build problems - sure there is a simple way to achieve this but the usual defconfig + diff do not seem to work.


lurch:
> Could not figure out how to trim the contents of kernelconfig_recovery... without getting build problems

Try having a play with `make_arch_linux-menuconfig.sh` - you should be able to do something like `./make_arch_linux-menuconfig.sh armv6` to run the menuconfig for the ARMv6 kernel, select the relevant options, and then afterwards run `./make_arch_linux-menuconfig.sh armv6 linux-savedefconfig` to save the defconfig in `buildroot/output/build/linux-b630db9dcb5c73367050a672823047b0466b16b9` which you'll then need to copy back to `buildroot/kernelconfig-recovery.armv6`. And then repeat the process all over again for the ARMv7 kernel (for Pi2)...

jumaka:
Have changed the kernel config files as per comments. Retested - all fine.

procount:
I have 2 suggestions:
1. Could you add a 'Rescan' button to the list of Access Points dialog? Sometimes the required access point is not listed the first time around and it is necessary to reboot to try again.
2. Could you add 2 cmdline parameters to specify the SSID and key? This would allow automatic  wifi selection, leading to automated OS installs (like silentinstall). This may require a couple of automatic rescans in case the Access Point is not seen during the first scan.
Otherwise, I like it! Apart from testing, what is involved in adding other wifi chipsets? Is it just a matter of enabling the appropriate driver and dongle firmware options?

jumaka:
On point 2, I can look at command line options, just not sure I have seen any arguments other than silentinstall, that does not seem to need a network. Is there a format for parameters?

In terms of additional cards, it is a change to the kernel config and a change to the buildroot config. So the process described earlier in this request for the kernel and a `make menuconfig` in the `buildroot` directory and selecting the relevant firmware files in `Target packages -> Hardware Handling -> Firmware -> Wifi Firmware`

procount:
On point 1. I had a look at the Qt APIs and I guess QNetworkConfigurationManager::updateConfigurations() would be an alternative to adding a rescan button, although you'll have to deal with some asynchronous signals.

For the parameters, you should follow the existing kernel cmdline parameter format in recovery.cndline, and convert it to a typical process cmdline parameter format in Init. Maybe something like ssid="my network" key="My pass" which gets converted to -ssid "My Network" -key "My Pass". 
You could look at my Pull Request #251 for another example. (I wonder if it is possible to use this to specify a static IP on the wifi too?)
I am aware of SSID and password having spaces in them but I'm not sure how they should be handled - I guess using the ubiquitous double quotes as in my example?

If I can get my hands on another wifi chipset, I'll try adding support for that as well.

XECDesign:
Sorry, we won't be adding this.

Issue #247, opened 2015-03-06
Cable network: manual config


Tabaluga:
Pleade add the Option to manually configure the cable network settings. I'm at work and there is no DHCP Server present. So I need to configure the network settings. Best regards.

XECDesign:
No plans to add this feature.

PR #251, opened 2015-03-12
Adding IP to cmdline.

### LINKED ISSUE #101: NOOBS v1.3b initialisation box doesn't disappear after failure to download distribution list
When booting up NOOBS v1.3b and attached to a network without direct internet access (ie. network proxy), the "Please wait while NOOBS initialises" dialog box doesn't disappear.

The dialog box "Error downloading distribution list from internet" appears, and when you close it, the dialog box mentioned above should, in my mind, terminate, but it doesn't. You can move it out of the way, and continue to install a local image.

If the network cable is not attached at boot, the "Please wait while NOOBS initialises" dialog box displays, then disappears after a few seconds, as it should.

procount:
Added an option to set static IP address parameters on recovery.cmdline, of the same form used by kernel NFS root parameters. i.e. ip=clientIP:serverIP:gatewayIP:netMask:HostName:Dev:autoconf. Solves #247.
serverIP and autoconf are not used.
This Pull request also fixes #101 by adding a timeout to the _qpd progress dialog


procount:
This PR worked well with eth0, but IIRC, it struggled a bit to set one of the other parameters for wifi (probably gateway address) so it may need some attention.

PR #252, opened 2015-03-22
Added complete Swedish translation

Mangero:
Had to add an if statement to languagedialog.cpp to enable automatic keyboard layout change (in the same way as for Japanese). Also updated the other translation files to have correct references to the source code. Built and tested on my RPi and everything seems to work (including installation of OSs).

Only thing is Arora (Online help) doesn't have Swedish locale but uses the language previously selected in NOOBS instead. With Swedish selected when entering recovery Arora uses Asturian language (which seems to be the default for some reason).


XECDesign:
Thanks. Could you please rebase the translation?

XECDesign:
Inactive

PR #267, opened 2015-06-17
Inform the user if more images are available

samnazarko:
OSMC was recently added to NOOBS (cheers Andrew).

Lately, we are finding a lot of users are downloading NOOBS Lite and are not plugging in a network cable and thus struggling to find OSMC in NOOBS.

This commit will let the user know of the potential to obtain more operating systems if they connect to the Internet. It makes sure *not* to display the dialog if there are no local images as well -- no point having two dialogs here.

Untested, as on holiday, but *should* work. Famous last words..

### DIFF
--- a/recovery/mainwindow.cpp
+++ b/recovery/mainwindow.cpp
@@ -142,6 +142,7 @@ void MainWindow::populate()
         _qpd->setWindowFlags(Qt::Window | Qt::CustomizeWindowHint | Qt::WindowTitleHint);
         _qpd->show();
         QTimer::singleShot(2000, this, SLOT(hideDialogIfNoNetwork()));
+        QTimer::singleShot(2000, this, SLOT(showMoreAvailableDialogIfNoNetwork()));
     }
 
     if (QFile::exists(SETTINGS_PARTITION))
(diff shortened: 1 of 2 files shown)


lurch:
I assume you meant "NOOBS" where you said "NOOBS Lite" above?

I'm not sure about this PR - users find popup dialogs annoying, so the fewer we have the better.
The instructions [here](https://www.raspberrypi.org/downloads/), [here](https://www.raspberrypi.org/documentation/installation/noobs.md) and [here](http://swag.raspberrypi.org/collections/frontpage/products/noobs-8gb-sd-card) all mention that additional OSes are available if you plug in a network connection.
I'll leave this open for a while to see if it attracts any further feedback, but I'll likely close it as invalid.

And for future reference, please ensure you fully test any PRs before submitting them :)

samnazarko:
>I assume you meant "NOOBS" where you said "NOOBS Lite" above?

Yes.

I understand it has been made clear on the Pi website but it does seem people are managing to avoid these notices.

NicoHood:
Is this still valid?

samnazarko:
We still sometimes see users asking why OSMC isn't in NOOBS or why they only see Raspbian. Andrew's comment about popups is also valid. 

There should be a nice way of telling the user that more images are indeed available. With the wpa-supplicant comment at top of tree at the moment, it would be nice to have users taking advantage of NOOBS' networking capabilities.

RaspberryPiFan:
@samnazarko You could also create this PR on PINN

PR #274, opened 2015-07-26
Add Asturian language

softastur:
Added files for Asturian translation


XECDesign:
I have merged #373, but that appears to have created a second 'Spanish' with these translations. Will hopefully be addressed in the next release.
Thank you.

procount:
QT uses the 2 letter language codes as defined in ISO639-1. Asturian is not listed in this set, but in the ISO 639-2 set of 3 letter language codes as 'ast'.
The es_XA code is 'made-up' and not valid.
In PINN, I therefore decided to use the 'ast' country code, but it needs an exception in https://github.com/procount/pinn/blob/2731e841bc3f6f14ca9031f7668202bba81ffbef/recovery/languagedialog.cpp#L84
I'm not sure what to do with the keyboard code though. It is a variant of the Spanish keyboard but not exactly the same (apparantly)

XECDesign:
Internationalisation is fun.

softastur:
We use the same keyboard layout as Spanish but if you press L and H keys + CTRL, you'll get U+1E24 (Ḥ) and U+1E36 (Ḷ) characters instead the U+0126 (Latin Capital Letter H With Stroke) and U+0141 (Ł) characters.

And yes, 'es_XA' is made up and its use must be avoided. Use 'ast' instead.

Issue #128, opened 2013-11-07
Menu navigation with TV remote (hdmi cec)


LodewijkSioen:
I'm using NOOBS as my multiboot system in my Rasberry. I have two OSses: OpenElec to use my pi as a media center and Raspian for toying around in Linux. I also have a bluetooth keyboard that I have paired in Raspian. Obiously, I cannot use this keyboard in the OS selection menu.

What I would love to do, is to use my TV remote control to select the OS I want to boot in. This is possible by using [HDMI CEC](http://en.wikipedia.org/wiki/HDMI#CEC). I can already use CEC to navigate in OpenElec, so there are [compatible libraries out there](http://elinux.org/CEC_%28Consumer_Electronics_Control%29_over_HDMI).

Is implementing CEC control possible in NOOBS?

atwright147:
Sorry to raise this issue from the ashes but couldn't libcec be setup as a compile option or couldn'y you offer two versions, one with and one without CEC?

calvarezn:
+1
The main reason why I use berryboot instead of noobs it's because of its CEC support.

Haris1977:
The problem in berryboot though is that you can not upgrade openelec ...:(

Ruffio:
@LodewijkSioen is this still an issue?

LodewijkSioen:
Well, there is no CEC control in NOOBS, so this is still an 'issue' (more a feature request than an issue).

However, if you have no plans to implement this, you can close this.

Sabcoll:
+1

Is there any way to edit noobs manually? I would like to add CEC support, even though its not implemented from scratch...

druidamix:
Raspberry is used a lot on televisions.
Have to use a mouse....
For me CEC Is essential.

is frustrating.
Noobs doesn't have cec
and berryboot can't update distros.

Alternative two sd's....

Anubi85:
+1

I use NOOBS for a multiboot system (OSMC, RetroPie, Raspbian)
When I  boot in OSMC or RetroPie I don't need a keyboard, so I think CEC is a good features to add to NOOBS :)

draxx31:
Hello everybody,

For me it is mandatory to have hdmi-ceci enabled in noobs. A lot of person use noobs to manage their different installation and in many case, they use their raspberry pi in the living room on the main TV with a media center (openelec for example).

It is not very useful to use each time a keyboard instead to use an all in one TV remote....

procount:
[...]

PR #280, opened 2015-09-22
Let users install CEC

Sabcoll:
I use Noobs to boot to OSMC or RecalboxOS on my TV.

Because the PI is connected to my TV, i have no mouse and or Keyboard connected and would love to control Noobs via CEC or via HTTP (for example as the Yatse APP in Kodi).

I already read about the license problems coming with CEC support from scratch, so i would suggest to make CEC a feature that could be installed manually.


lurch:
Closing as a duplicate of #128

MartinX3:
Why is that a duplicate?
Noobs got CEC integration?

PR #281, opened 2015-09-27
Added Polish translation (recovery)

clansfear:
Complete /recovery PL translation, tested numerous times.


XECDesign:
See #373
Thank you

PR #284, opened 2015-09-30
Fix parted size invocation

procount:
The use of parted to resize (shrink) the first partition is incorrect. It uses the Start of the rescue partition (in sectors) and the Size of the rescue partition (in MB). It should use the Start and End location of the partition when resizing. I modified the invocation to use the End of the rescue partition (in sectors).

This isn't going to make a lot of difference, since the start of the rescue partition is pretty close to the start of the disk and 4MiB isn't going to eat much into the 100MB buffer that is added. So this change is more for correctness and understandability.

### DIFF
--- a/recovery/initdrivethread.cpp
+++ b/recovery/initdrivethread.cpp
@@ -253,7 +253,11 @@ bool InitDriveThread::method_resizePartitions()
         newStartOfRescuePartition = 8192; /* 4 MiB */
     }
 
-    QString cmd = "/usr/sbin/parted --script /dev/mmcblk0 resize 1 "+QString::number(newStartOfRescuePartition)+"s "+QString::number(newSizeOfRescuePartition)+"M";
+    /* Parted needs start & End (not start & size) */
+    /* Convert size from MiB to sectors & add to start to get end position*/
+    int newEndOfRescuePartition =  newStartOfRescuePartition + (newSizeOfRescuePartition * 2048);
+
+    QString cmd = "/usr/sbin/parted --script /dev/mmcblk0 resize 1 "+QString::number(newStartOfRescuePartition)+"s "+QString::number(newEndOfRescuePartition)+"s";
     qDebug() << "Executing" << cmd;
     QProcess p;
     p.setProcessChannelMode(p.MergedChannels);


lurch:
Yeah, I'd also noticed this before, but didn't think it was worth fixing ;-)

maxnet:
Be careful that specifying sectors tells parted to resize the file system to exactly that amounts of sectors, not one more, not one less.
If it is happens to be not a valid exact amount of sectors for a FAT file system, it will fail with an unable to satisfy constraints error message.

procount:
@maxnet - thanks for that hint! I would still like to change the semantics so that it reads correctly. So if I convert the units from sectors to MB, I assume Parted will adjust the partition size to be the exact number of sectors to satisfy the FAT constraints? I note that the partition size is re-read after the resizing - I assume for this very reason.

maxnet:
Yes

M = roughly that amount in 1,000,000 bytes, adjusted to satisfy constraints
s and MiB = exact in sectors or MiB


While you are at, note that the newSizeOfRescuePartition calculation is slightly wrong.

``
int newSizeOfRescuePartition = sizeofBootFilesInKB()/1000 + 100;
``

The amount returned by sizeofBootFilesInKB() is actually in KiB, so if you want it proper, it should be something along the lines of:

``
int newSizeOfRescuePartition = sizeofBootFilesInKB()*1.024/1000 + 100;
``

procount:
Thanks. 

Whilst I have your attention on parted - Noobs is still using v2.3. Do you think upgrading to v2.4 or v3.2 might fix the 0MB problem? I can't reproduce the problem myself, but maybe I can put some test versions out there for those affected to try it on their SD cards?

Also I note that the RISC partition MUST start at 1760MB and is of a fixed size. Is this because the start is hardcoded into RISCOS_BOOT.BIN ? Was this a balance to allow sufficient space for the compressed distros in the recovery partition, yet still fit on a 2GB card? When using NOOBS lite, this leaves a ~ 1.6G hole that is unavailable. Could this be used as a Data partition? Does RISC_OS need to be in partitions 5&6 (maybe because the risc boot partition references p6?)? I suppose P1 could be extended to fill this space when RISC is installed?

maxnet:
As of v2.4 parted no longer supports resizing file systems at all, the authors gave up on that...

Regarding users complaining NOOBS reports 0 MB.
I suspect those users receive an error first, like unable to resize FAT partition due to the two FATs being corrupt and not matching.
If they click away the error, NOOBS will not shut down the system but continue. And indeed will report 0 MB available because it is unable to mount the ext4 partition, given it was not able to create that one either in that case.
The error flow could be improved to not allow an user to continue after an error.
But it will not solve the actual problem that parted was unable to resize the partition.
Maybe running fsck before the resize may clear some errors, but it is not a great solution either, as it may not be just the file system structures that are corrupt, there could be data corrupt/missing as well, especially if the user removed the SD card from his computer prematurely without safely removing it.

RiscOS does not understand partitions at all, and expects things at hard coded offsets.

procount:
I thought 2.4 was the last version that did support resizing. They took it all out in v3, then put it back better in v3.2? I already upgraded my NOOBS to use Parted 2.4 and it still worked but I don't know if it improved anything, because 2.3 works for me too.

I'd like to help out with the 0MB problem. Maybe my first step, then, is to improve the error flow and logging to identify the root cause....It all seems very odd to me....

maxnet:
>They took it all out in v3, then put it back better in v3.2?

No, never returned.
At least not in parted, the command-line application.
libparted is a different story, but linking to that has licensing implications.

procount:
I bow to your greater knowledge - I have just looked at the release notes so far from [url]http://savannah.gnu.org/forum/forum.php?forum_id=8042[/url] where it is mentioned they added 'resizepart' but I've not tried it out. I will steer clear of libparted.

maxnet:
>where it is mentioned they added 'resizepart' 

```
              resizepart partition end
                     Change  the  end position of partition.  Note that this does not modify any filesystem present
                     in the partition.
```

That's resizing a partition (altering the partition table) versus resizing a file system (messing with FAT structures)

procount:
Thanks - now I won't even bother looking further at v3.2

PR #323, opened 2016-03-11
Add bcm2710-rpi-3-b.dtb (fixes WiFi support for Raspberry Pi 3)

matthuisman:


procount:
I tried to implement this fix, but I can't get it to work.
I cleaned out the output/build/rpi-firmware-3442862c10fab68c2e88d660d2e69c143bb1f00c folder and ran ./BUILDME/sh (with and without skip--kernel-rebuild), but the 3B dtb file cannot be found and breaks the build. Interestingly, the DTB files that do exist are all dated Feb 9th, so they are not getting rebuilt.
I seem to recall there is a special make command to make the DTBs. Any idea what it is?

maxnet:
```./BUILDME.sh update-firmware```

procount:
Ahh, it needs an updated firmware since the DTBs are pre-compiled. That seems to have fixed it.

DTS files are in the linux repository, so would updating linux to 02ce9572cc77c65f49086bbc4281233bd3fa48b7 have achieved the same thing? (I think this is where the special make target comes in that I was thinking of?)

Or is it that NOOBS doesn't build the DTBs from linux, but just gets the rpi-firmware prebuilt ones for build speed reasons?

maxnet:
>Or is it that NOOBS doesn't build the DTBs from linux, but just gets the rpi-firmware prebuilt ones for 
>build speed reasons?

We used to build DTBs from source briefly, but I disabled that again because building overlays is broken in buildroot.

See: http://lists.busybox.net/pipermail/buildroot/2015-June/131101.html


While overlays are normally not strictly necessary, I figured at least the rpi-ft5406 one might be for the official touch screen to work...

procount:
>  I figured at least the rpi-ft5406 one might be for the official touch screen to work...

Indeed! Don't take that one out! ;-) Thanks for the info.

lurch:
Fixed by https://github.com/raspberrypi/noobs/commit/b7d838d5b0cc2db436c5ca1265d9999bb5b53b85 and released in NOOBS v1.9

PR #326, opened 2016-03-13
Make timeout configurable in config.h

fonic:
I think the timeout is something that should be configurable. Here's the tiny patch for it.

EDIT:
Sorry for the translation file being part of the pull request. I'm new to GitHub, still tricks me from time to time. Can be safely merged as the translation file is the same as in my other pull request for the translation.


XECDesign:
Sorry, we won't be adding this.

fonic:
Why not? If the reason is the translation files, I can create a new, clean pull request.

XECDesign:
Alright, it's optional and harmless, so may as well add it. Send a clean PR.

fonic:
Will do!

PR #327, opened 2016-03-13
Updated German translation (complete)

fonic:
While waiting for NOOBS to build, I used the time to update the German translation.


XECDesign:
See #373
Thank you

PR #328, opened 2016-03-14
Use BACKGROUND_COLOR instead of hardcoded Qt::white

fonic:
This makes it much easier to alter NOOBS to a black (transparent) background by only adjusting config.h.

I use it like that since version 1.4 since I think it's easier on the eyes...


XECDesign:
Thanks

Issue #158, opened 2013-12-27
Can't build NOOBS per instructions in NOOBS readme


IvanExpert:
Using a completely fresh install of Ubuntu Server 13.10 32-bit (including both before and after apt-get upgrade and apt-get dist-upgrade), "sudo ./BUILDME.sh" plows along fine for about an hour, and then dies at:

/home/user1/noobs/buildroot/output/host/usr/bin/moc -DQT_NO_DEBUG -DQT_GUI_LIB -DQT_NETWORK_LIB -DQT_CORE_LIB -DQT_SHARED -I../../host/usr/arm-buildroot-linux-uclibcgnueabi/sysroot/usr/mkspecs/qws/linux-arm-g++ -I. -I../../host/usr/arm-buildroot-linux-uclibcgnueabi/sysroot/usr/include/QtCore -I../../host/usr/arm-buildroot-linux-uclibcgnueabi/sysroot/usr/include/QtNetwork -I../../host/usr/arm-buildroot-linux-uclibcgnueabi/sysroot/usr/include/QtGui -I../../host/usr/arm-buildroot-linux-uclibcgnueabi/sysroot/usr/include -I. -I. bootselectiondialog.h -o moc_bootselectiondialog.cpp
/home/user1/noobs/buildroot/output/host/usr/bin/rcc -name icons icons.qrc -o qrc_icons.cpp
RCC: Error in 'icons.qrc': Cannot find file 'translation_nl.qm'
make[1]: **\* [qrc_icons.cpp] Error 1
make[1]: **\* Waiting for unfinished jobs....
make[1]: Leaving directory `/home/user1/noobs/buildroot/output/build/recovery-1.0'
make: **\* [/home/user1/noobs/buildroot/output/build/recovery-1.0/.stamp_built] Error 2

This is consistently reproducible for me. I'm using VirtualBox 4.3.6 on a Mac as the Ubuntu host machine. Anything you want me to try doing differently, or can I provide any more detail?

procount:
`sudo apt-get install libqt4-dev` allows lrelease to be built and is stored in buildroot/host/usr/bin. This allows a first build of noobs/recovery. However, lupdate is not built by buildroot, so the build could subsequently fail depending on what modifications are made.
To add lupdate to the build, qt.mak needs updating and projects.pro needs patching.

Here is the diff file for the qt.mak changes:
-----------------------------------------------
[...]

PR #329, opened 2016-03-16
Updated README.md to v1.8 user features

procount:
I have updated the NOOBS README.md file to include the user visible changes and new features up to v1.8 to help people make use of and understand the latest version

I have not documented any changes that were made "under the hood", like partition layouts, use of p7zip or NTFS support etc as I didn't think they were relevant to the user documentation.

I have not updated any screenshots, so this may still need to be done as they are still showing v1.3.

(To view theses changes, I found it best to switch the github diff viewer to "rich" mode otherwise it shows the whole file has changed. Not sure why that is - line endings maybe?)


NicoHood:
It would be nice to get the 1.8 source too. I think that is 1.6 or something like that.

procount:
If you followed my discussion with maxnet, you'll know that this repository source is actually up to date with v1.8 apart from the change of version number from v1.7 to v1.8 in config.h (and the ommision of the RPI3B DTB file). Apparantly that is part of the binary release done by someone else, and not part of the source release. In any case it does cause some confusion when the last release omits the version number change because we don't know if there is anything else missing. Hopefully they will remember when v1.9 is released :-)

XECDesign:
Looks like a lot of good changes, thank you.
@maxnet, okay to merge?

maxnet:
```
#### What if my built in Wifi does not work

If you have a RPi 3B model and the built-in Wifi does not work with NOOBS, check if you have the bcm2710-rpi-3-b.dtb file in the NOOBS root partition. If it does not exist, copy a version from the latest Raspbian release.
```

No longer necessary in 1.9

===

```
-`sudo apt-get install build-essential rsync texinfo libncurses-dev whois unzip`
 +`sudo apt-get install build-essential rsync texinfo libncurses-dev whois unzip bc qt4_linguistic_tools`
```

Rather than expand the list of requirements, I wonder what needs bc? (and whois and texinfo?)

lurch:
I think `qt4_linguistic_tools` should actually be `qt4-linguist-tools` ?

procount:
Good catches! I will update.
Is there anything else in v1.9 I could document? (a source code update would be nice ;-) )

procount:
`qt4-linguist-tools` is definitely required. Can't remember why I noted down `bc` as well - it was so long ago. The rest were original requirements and I have no idea if any can be discarded.

lurch:
IIRC there's some part of the kernel-build that requires `bc`, `whois` is required because something needs the `mkpasswd` binary the package also includes, and `texinfo` is needed because some buildroot package (I can't remember which) tries to build its documentation even though the `--disable-docs` flag (or equivalent) is passed to the package's `configure`.

But it's been a while since I looked at NOOBS, so I could be remembering wrong!

Also, I don't think the "RPF Broadcom wifi dongles" language is very clear (perhaps a link to https://www.raspberrypi.org/products/usb-wifi-dongle/ would be useful), as this is "official documentation" then perhaps "RPi 3B" should be spelled out in full as "Raspberry Pi 3 Model B", other usages of "RPi" probably ought to be "Pi" to match the rest of the README, and I think "RPF Touchscreen" ought to match the name used on https://www.raspberrypi.org/products/raspberry-pi-touch-display/

P.S. `qt4-linguist-tools` is only needed because I never got round to properly fixing #158

maxnet:
You are right, ``bc`` is indeed required as well to build a kernel nowadays.
buildroot does check if it is installed, but seems they forgot to update the documentation.

https://git.busybox.net/buildroot/commit/support/dependencies/dependencies.sh?id=7cbebbc48c26ddfb18da23ffba356744a5592ea6

==

Would expect mkpasswd to be build by the host-mkpasswd buildroot package though.
https://github.com/raspberrypi/noobs/tree/e63c7ee0a067c45355698e4772bba071a9e87492/buildroot/package/mkpasswd

procount:
I've updated the pull request to included all comments. Let me know if I missed any (or any 1.9 changes to be added)
(Dunno how those file permissions got changed *shrug*)

procount:
I have created some updated screenshots to go with the updated README.md. using v1.9

os_installed
os_selected
config_editor

I also added a screenshot for the wifi selector (wifi_selector.png) But it needs the following link added under the "Wifi Networks" heading:

`alt text`


wifi_selector

lurch:
I've got no idea why the screenshots are hosted on the downloads server (I suspect the original reason may be lost in the mists of time). Anyone have any objections to them just being hosted directly on github? (with README.md obviously being modified to suit)

Hmmm, and as there are currently fewer OSes available (in NOOBS) on a Pi2 / Pi3 than a Pi1, maybe it would look better if the screenshots were instead taken on a Model B+ ?

procount:
Would that give Pi2 / Pi3 users false expectations, leading to complaints? 
Whereas earlier model users would currently be pleasantly surprised?
Maybe worthy of an additional note to say something like: "Additional OSes may be made available over time and may differ between Pi models?

lurch:
> Maybe worthy of an additional note to say something like: "Additional OSes may be made available over time and may differ between Pi models"?

Sounds good, please add it to your readme PR. And I guess I ought to add a similar note to https://www.raspberrypi.org/documentation/installation/noobs.md too!

procount:
> Anyone have any objections to them just being hosted directly on github?

I'm just musing....
1. The links would not work until the images are pushed up to the public github repo. (Well I guess that's an issue no matter where they were hosted)
2. If you develop on another github repo and link to images on there, the links would need to change to point to the public repo before pushing/releasing. If you don't, it may not be obvious that the public repo is now linking to images on another repo.. (unless they are now inaccessible).
3. Is there any way to use "relative" links to images on the github, so they automatically link to the same repo the README file is on? So instead of `github.com/raspberrypi/noobs/raw/master/screenshots/os_installed.png` can we just use `raw/master/screenshots/os_installed.png` or something?

procount:
I have pushed another update that copies the screenshot images to the github (using a relative URL), adds the note about available OSes under 2 of the images, and added a screenshot of the wifi dialog.

I haven't squashed this set of commits yet in case someone disagrees with any of the changes, but when you are happy I can squash them.

procount:
As there have been no more comments, I have squashed my readme branch so it is all ready for integrating.

NicoHood:
https://github.com/procount/noobs/tree/readme#how-to-format-an-sd-card-as-fat

I never know if fat means fat16 or fat 32. And there was a 3rd name like msfat or something like that. Maybe you should clarify this (for the linux section, dont know about micro$$$$)

lurch:
> I never know if fat means fat16 or fat 32.

The Raspberry Pi's firmware (and hence NOOBS as well) can boot from both FAT16 and FAT32 - so it doesn't matter which you use. It's only exFAT that isn't supported https://www.raspberrypi.org/documentation/installation/sdxc_formatting.md

procount:
Does anyone want to merge this Pull request before it gets out of date, or are you waiting for the release of v1.10?

XECDesign:
I haven't checked all of it, but it certainly is better than what's there already in any case. Many thanks.

procount:
Woo-hoo!
Many thanks.

Issue #168, opened 2014-01-12
[feature request] Default OS and Timeout


mhtrinh:
I would like to suggest those feature for NOOBS: 
- set a default OS to boot (and still keep the feature boot to OS from last time)
- set timeout for default OS

Ruffio:
@mhtrinh is this still an issue?

mhtrinh:
Because NOOBS lack this feature, I end up using another way to do it. If some day this feature end up in Noobs, then I will probably use it. 
But it seems I'm the only one who need this ...

robgjansen:
I would very much like the ability to make the default OS sticky, rather than always automatically updating the default OS to the last-selected OS. I like the option suggested here:

    default_partition_to_boot=last # boot to last selected OS
    default_partition_to_boot=5 # boot to partition 5, no matter the last selected OS

The use case is when the user is dual-booting where one installed OS does not require a graphical interface (e.g., raspbian lite, and/or pidora configured with runlevel 3) and one installed OS does require a graphical interface (e.g., OpenELEC, and/or RetroPie).

When the non-graphical OS is the main OS and primarily accessed through a serial cable, it gets very annoying to boot back into that after, e.g., connecting to the TV to play a game in RetroPie. Currently, the user would need to boot back into raspbian from the TV to make sure the default is updated before going back to the serial cable method, or otherwise figure out how to get to the correct partition to update the setting manually.

Concerning the issue raised about booting into a non-existent partition: I believe that is a rare condition and should be testable programmatically. If the default partition doesn't exit, we can go back to the "last" setting again.

robgjansen:
One way to implement this:

`default_partition_to_boot=5` would boot partition 5, and auto-update the setting to the last-selected OS (current behavior), and then `default_partition_to_boot=-5` would also boot partition 5, but there would be no auto-update of the setting.

To handle the "partition does not exist problem", in that case we can boot into recovery, since that is always in partition 1.

procount:
I have implemented my take on this requirement in PINN.
Please look in https://sourceforge.net/projects/pinn/files/testing/ and use pinn-212a.zip in the same way as noobs-lite.
By adding 'bootmenutimeout=X' to recovery.cmdline you can change the OS boot selection menu timeout to X seconds.
[...]

PR #330, opened 2016-03-19
Issue168

robgjansen:
This implements parts of #168. This was not tested and a build was not attempted.


XECDesign:
Sorry, no plans to add this right now.

Issue #336, opened 2016-04-29
problems with DHCP opt 61 on Raspi3


mam59:
Before I begin: english is not my natural language, it maybe a bit hard for me the rather complex problem I have run into recently.

The problem could briefly be titled "**NOOBs on a raspberry 3 cannot get a dhcp lease from a hardened microsoft dhcp server"** (other vendors may be affected too, but are not tested)
Let us take a look into the request packet the NOOBs (1.9) sends out at boot time:
![raspi3dhcp](https://cloud.githubusercontent.com/assets/17042691/14911215/2fc5c998-0df4-11e6-8662-4e02cad292b4.jpg)
We see that raspy with the ethernet mac address of ... AC-39-B0 sends out a request with an client id of (combined) "ebac39b0br27ebf96ce5", the first part is taken from the LAN card (Raspberry vendor stripped), the last part is the complete MAC from the WLAN card onboard.
This is a very nice and valid clientid for **IPV6!**, but it kills V4 if the server is configured to run in a mode with either reservations defined or with a mac filter.
If the microsoft server finds an option 61 in a packet it takes this id in favor to the mac address the packet originated from!
And, in the V4 part of the Server you have no chance to define a filter with such an id, only plain mac addresses are available!
So the admin has no chance to add this raspi to the list of allowed clients.
To let him in, he has to disable any filter mode of the server, then option 61 is ignored and the client is given a lease.
But even if he lets him in this way and then wants to freeze the given lease into a permanent reservation, it fails because the invalid (for V4) client id.

It seems to me like somebody has written the docs too deeply, NOOBs could safely drop this option 61 and everything will be fine again..

Sometimes doing less is giving more...

procount:
My guess is that changing the `duid` option to `clientid` in  `/etc/dhcpcd.conf` (the same as in Raspbian) will fix it, but I've not tested it yet. Any comments @maxnet, @XECDesign, @lurch?

mam59:
Nice Idea but I'm afraid, it wont help much. DUIDs are also defined for V6 only.
The clientid for V4 should simply be Hardware-Type + Mac Address.
I still not have found the source for the DHClient program to look up, what it does internally to construct the client id. Isn't it within the NOOBs master tree somewhere?

Funny Joke: if NOOBs would send out this kind of "wrong" packet here in our LAN, just on a V6 socket, it would succeed and find a nice Server that would reply to it. :-)

maxnet:
[...]

PR #337, opened 2016-05-01
dhcpcd uses clientid, not duid

procount:
Uses the `dhcpcd.conf` file copied from Raspbian that uses `clientid` option instead of `duid`. 
Fixes issue #336.
I moved the fix to the dhcpcd package instead of putting it in the recovery package.
An alternative implementation would be to patch the `dhcpcd.conf` file instead of using a completely new version, which would probably make it more robust to future changes.

### DIFF
--- a/buildroot/package/dhcpcd/dhcpcd.mk
+++ b/buildroot/package/dhcpcd/dhcpcd.mk
@@ -36,9 +36,11 @@ endef
 
 define DHCPCD_INSTALL_TARGET_CMDS
 	$(TARGET_MAKE_ENV) $(MAKE) -C $(@D) install DESTDIR=$(TARGET_DIR)
+	$(INSTALL) -D -m 0664 package/dhcpcd/dhcpcd.conf $(TARGET_DIR)/etc/dhcpcd.conf
 endef
 
 # NOTE: Even though this package has a configure script, it is not generated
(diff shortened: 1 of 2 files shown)


buildroot/package/dhcpcd/dhcpcd.mk
```diff
@@ -36,9 +36,11 @@ endef
 
 define DHCPCD_INSTALL_TARGET_CMDS
 	$(TARGET_MAKE_ENV) $(MAKE) -C $(@D) install DESTDIR=$(TARGET_DIR)
+	$(INSTALL) -D -m 0664 package/recovery/dhcpcd.conf $(TARGET_DIR)/etc/dhcpcd.conf
```
lurch: Is this right? The other file in this PR seems to be in package/dhcpcd/... rather than package/recovery/... ?

procount:
Doh! Of course it worked on my machine but won't on anyone elses's! Will fix it. Thanks for spotting.

procount:
Bug fixed, branch rebased and pushed.

XECDesign:
@maxnet, does this look good to merge?

maxnet:
Looks good to me.

XECDesign:
Thanks

PR #344, opened 2016-06-07
Add note about sparse files

Ealdwulf:
The `tar` implementation used doesn't unpack sparse files correctly, so add a note about this ( we were using bsdtar to generate the archive, and a random file (usr/lib/locale/locale-archive) was for some reason sparse, so it wasn't unpacked currently).


XECDesign:
Is this still an issue given that we use bsdtar to unpack now?

PR #357, opened 2016-09-27
Finished Turkish translation

diorray:
Completely translated to Turkish, added/modified necessary files and tested.


XECDesign:
See #373
Thank you

PR #183, opened 2014-02-25
Added Spanish translation.

amorellgarcia:

PR #362, opened 2016-12-14
Italian GUI translation

bovirus:
This the Italian GUI translation.
Pelase rebuild GUI adding Italian. Thanks.

### DIFF
--- a/recovery/translation_it.ts
+++ b/recovery/translation_it.ts
@@ -1,0 +1,793 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE TS>
+<TS version="2.1" language="it" sourcelanguage="en">
+<context>
+    <name>BootSelectionDialog</name>
+    <message>
+        <location filename="bootselectiondialog.ui" line="19"/>
+        <location filename="bootselectiondialog.cpp" line="249"/>
+        <source>Select OS to boot</source>
+        <translation>Seleziona OS da avviare</translation>
+    </message>
+    <message>
+        <location filename="bootselectiondialog.cpp" line="44"/>
+        <source>Cannot display boot menu</source>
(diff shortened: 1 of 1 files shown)


Ruffio:
@lurch Is it possible to get a general statement about if you accept translations or not? Currently there are 20 translation PR waiting to be merged, some quite old. If you don't accept translation PR then please reject them, else process them and get them merged if they are good enough. There is no point in people use time to make the PR if you are not going to process them and they are just 'stalling'. That is just a waste of time and good energy...

Ruffio:
@bovirus I don't have any preferences for languages, but I'm sadden to see that people use time to translate and make PR and then their work is just ignored by the foundation. I think that it is a very arrogant attitude to have. At least the foundation could write that they are not interested in the PRs and then reject them, then people would know that they shouldn't use time to make a PR. IMHO one should be thankful for free work that can improve your product. At least the foundation could reject the PRs if they are not going to be merged, else write what should be fixed in order for them to get merged. Just ignoring other peoples work is so disrespectful.

Ruffio:
@ghollingworth please be a little more serious about this instead of being arrogant and trying to push this away. The track record shows that you indeed ignore the translations, so write it somewhere and reject them instead of just let them be hanging. It is easy to retranslate the translation to see if it make sense even if it is a language that you don't understand yourself. Else write that you will wait for someone else to confirm the translation. Just do something instead of doing nothing.

Take a look at this: https://github.com/raspberrypi/noobs/pull/183 
On 8 Jan 2015 @lurch had confirmation of the translation from two other persons (@Gioyik , @avegao) and @lurch wrote that he would merged it in with the next release of NOOB...

Based on your statement above and issue #183 I can conclude that it takes you guys longer than a year to think twice. Really?

XECDesign:
Thanks

PR #377, opened 2017-01-19
Copy ssh for Raspbian

procount:
This patch will fix the NOOBS issue created by disabling ssh by default in Raspbian.
It allows a user to put a blank ssh file on the recovery partition, which NOOBS will copy to the /boot partition, but only for Raspbian. The first IF statement could be removed to allow it to work with all OSes.

Without this it is difficult to do a headless Raspbian install using NOOBS and get ssh working, although I did post some other workarounds for this on the forums. This makes the process the same for both a NOOBS and an image install of Raspbian.

### DIFF
--- a/recovery/multiimagewritethread.cpp
+++ b/recovery/multiimagewritethread.cpp
@@ -491,6 +491,17 @@ bool MultiImageWriteThread::processImage(OsInfo *image)
     emit statusUpdate(tr("%1: Saving display mode to config.txt").arg(os_name));
     patchConfigTxt();
 
+    /* enable SSH
+     * If user supplied a ssh on the FAT partition copy that to partition1 (boot)
+     */
+    if (os_name == RECOMMENDED_IMAGE)
+    {
+        if (QFile::exists("/mnt/ssh"))
+        {
+            QFile::copy("/mnt/ssh", "/mnt2/ssh");
+        }
+    }
+
     /* Partition setup script can either reside in the image folder
      * or inside the boot partition tarball */
     QString postInstallScript = image->folder()+"/partition_setup.sh";


maxnet:
But does Raspbian specific stuff belong in NOOBS itself?

Perhaps better to have that in the Raspbian post-installation script?

XECDesign:
Interesting solution. However, it looks like it misses raspbian lite and ssh.txt. I intended to do what maxnet suggested for the next raspbian release.

procount:
Originally, I didn't include the RECOMMENDED_IMAGE so it would work for all,OSes including lite and I missed the fact that ssh.txt was also allowed, but that is easily added.

I already posted a solution in the forum about the partition_setup.sh post-installation script, but I figured this way was more consistent and easier for a newb to apply. How were you thinking of getting the user to enable the option? Especially if you are going to download Raspbian instead of install it from the SD card itself - the partition_setup.sh script is not available in that case. It needs to work for a headless install too.

XECDesign:
I don't quite follow the question. What's the problem with using the same mechanism you have above, but in the partition_setup.sh script? If the file is there, copy it to the target boot partition.

procount:
Oh sorry - I was on a totally different planet for a moment and got the wrong end of the stick. Yes of course, that's totally compatible ! :+1:  (What was I thinking? Doh!) Can I just delete this whole PR?

XECDesign:
I am afraid github does not have the option of deleting a PR. Let's just all pretend it never happened.

procount:
Lol! Anyway, glad to hear it is being addressed. How are the USB boot mods going?

XECDesign:
I am not involved in development, but last time I heard it mentioned, it was in a "that thing that has been done already" context. Probably a question for @maxnet. I am guessing it's waiting for usb boot to get past the experimental stage.

Issue #383, opened 2017-02-03
Touchscreen tap stopped working?


procount:
Is it just me, or has the touchscreen tap just stopped working to enter the NOOBS mainwindow (v2.1)?
Once the mainwindow has been entered, the touchscreen works fine, but the initial tap has stopped working for me.

procount:
Ahh, it's just me. BUT it seems necessary to click on the central image. It doesn't wok if you click on the background. Should we add something to the README to that effect?

XECDesign:
I am not sure how many people this would actually affect, but if it happened to you, I suppose there are probably others too. Perhaps a note in README would be okay.

PR #388, opened 2017-03-01
Add note about touchscreen tap

procount:
Minor amendment to clarify where to tap the touchscreen instead of using the shift key. See #383


XECDesign:
Looks good, thanks.

PR #389, opened 2017-03-03
select OS to boot based on state of GPIO pin(s)

RenardFox:
When running a Pi with more than one OS installed, it would be particularly useful to select the OS to boot each time using a similarly readily-configurable-headless method, such as the state of a GPIO pin.

(This particular suggested method has the advantage that, when the Pi is used with multiple GPIO-using projects, it is possible for the add-on hardware to be configured to automatically choose which OS to boot by setting the startup state of the relevant GPIO pin.)

Thanks


XECDesign:
Commits 98, Files changed 5,334, "This branch has conflicts that must be resolved"?

PR #411, opened 2017-06-09
add zh_CN translation

wolfg1969:


XECDesign:
With this (plus a fix fixes added), zh_TW and zh_CN both show up as "中文". Would it make sense to have "普通话" and "國語"? If not, what's the correct way to differentiate them?

@wolfg1969 
@s8321414

s8321414:
For me, I think 中文（臺灣）for zh_TW is better, don't know which name does @wolfg1969 want for zh_CN?

wolfg1969:
How about 简体中文 (Simplified) for zh_CN and 繁體中文 (Traditional) for zh_TW?

Lin-Buo-Ren:
I support 中文（台灣）

wolfg1969:
简体中文 - Chinese, Simplified 
繁體中文（台灣）- Chinese, Traditional (Taiwan)

Lin-Buo-Ren:
Much better

XECDesign:
Looks like Qt supports 中文 (简体) and  繁體中文. I'll have to go with those unless I want to make significant changes to how translations work. Or we could add an additional string for the language name in the translations file to give more flexibility.

Artoria2e5:
Appearing for no reason, but in principle I am against labeling zh-CN as the sole representative of zh-Hans. A good bunch of zh-Hans users are in SG and MY. Good Qt bug report material.

XECDesign:
Have to pick a flag to represent the language and since most Hans users are probably in PRC, I don't think there's any way around that.

XECDesign:
This is probably the best we can do right now with the existing mechanism.
screenshot

Lin-Buo-Ren:
Anyone now existing Qt bug report regarding this issue?

Artoria2e5:
Nah, I've got no objection to the flags: CN flag should be used for zh_CN, obviously.

It's just the use of 简/繁 -- script labels -- as national/locale labels that bothers me.

XECDesign:
> Anyone know existing Qt bug report regarding this issue?

What would be the bug exactly? Right now we are using nativeLanguageName(), which returns "中文" for both zh_TW and zh_CN locales. I don't think it's wrong in the context of Qt - it correctly reports the language name.

> It's just the use of 简/繁 -- script labels -- as national/locale labels that bothers me.

I don't follow. Since localisation is complex and you can have multiple languages, scripts, countries, doesn't it make sense? Given than China has multiple scripts, wouldn't you want to specify which script this translation is using. Or is there some nuance I am missing here?

Artoria2e5:
> What would be the bug exactly? 

1. The formatting for 中文 (简体) and 繁體中文 is not self-consistent.
2. Should be 中文（中国）and 中文（臺灣）for corresponding to `zh-CN`/`TW`. See below.

> Right now we are using nativeLanguageName(), which returns "中文" for both zh_TW and zh_CN locales.

Localization is not just putting it into a language; it's about putting stuff into a locale. This is why you have all those country suffixes after the two-letter language codes. If Qt does this thing correctly for `fr-FR`/`fr-CA` and `en-US`/`en-GB`, there's no reason it shouldn't get it right for Chinese locales. Since Qt's doc gives an example for `de_CH` ("Schwiizertüütsch"), it's obvious that we have found a Qt bug here.

> Given than China has multiple scripts

China doesn't have multiple functional scripts for writing the Chinese/Mandarin language. `zh-CN` as defined in the FOSS world corresponds to the current official variant of Mandarin (`cmn`) promoted by the Chinese government, which is written in a simplified script (`Hans`). In the very elaborate web language code world we call it `zh-Hans-CN`. The same follows for `zh-Hant-TW`.

The reason for not using script labels for actual country-bound locales is simple: neither `Hans` nor `Hant` is confined to a single region. Chinese as written in Singapore and Malaysia also uses the `Hans` script, but may take on local numbering formats and stuff. Mandarin Chinese as written in Hong Kong and Macau uses the `Hant` script, with its technological vocabulary dramatically different from that of Taiwan.

You don't make the label for `en-US` "English (Americanized spelling)", or Canadians would be upset because they don't get to wear their *tuque*.

XECDesign:
Not having a whole lot of time to spend on this, I'm going to leave it as is. Sensible pull requests to address this properly are welcome.

PR #413, opened 2017-06-14
Update Portuguese translation by igorruckert

bovirus:
Update Portoguese translation by igorruckert - Please add it as reference as translator


XECDesign:
Please run pre-commit-translation-update-hook.sh before commiting translations.

XECDesign:
Thanks

PR #418, opened 2017-06-22
USB Disk Sizes

procount:
Now that NOOBS can support installation to USB MSDs, it opens the door to some very large capacity devices. A recent user of PINN reported an inability to install 2 or 3 OSes to a 2TB drive. The same errors occurred in NOOBS consisting of the display of negative sizes and errors creating partitions.

A lot of the variables used  to calculate disk offset, sizes and sector numbers are just integers and are too small for these large capacity drives. So I have converted all of them to qint64 and used the .toULongLong() function where appropriate as a proposal.

I haven't got a 2TB disk to test it on myself, but the user reported that he was able to install the OSes with the changes without errors. Therefore you might like to review all of the changes before implementing, but nevertheless support for USB drives >1TB should probably be included now.

### DIFF
--- a/recovery/progressslideshowdialog.cpp
+++ b/recovery/progressslideshowdialog.cpp
@@ -132,13 +132,17 @@ void ProgressSlideshowDialog::changeDrive(const QString &drive)
 
 void ProgressSlideshowDialog::setMaximum(qint64 bytes)
 {
+    /* restrict to size of 1TB since the progressbar expects an int32 */
+    /* to prevent overflow */
+    if (bytes > 1099511627775LL) /* == 2147483648 * 512 -1*/
+        bytes = 1099511627775LL;
     _maxSectors = bytes/512;
     ui->progressBar->setMaximum(_maxSectors);
 }
 
 void ProgressSlideshowDialog::updateIOstats()
 {
-    qint64 sectors = sectorsWritten()-_sectorsStart;
+    uint sectors = sectorsWritten()-_sectorsStart;
 
     double sectorsPerSec = sectors * 1000.0 / _t1.elapsed();
     if (_maxSectors)
@@ -155,7 +159,7 @@ void ProgressSlideshowDialog::updateIOstats()
     }
 }
 
-qint64 ProgressSlideshowDialog::sectorsWritten()
+uint ProgressSlideshowDialog::sectorsWritten()
 {
     /* Poll kernel counters to get number of bytes written
      *
@@ -177,6 +181,8 @@ qint64 ProgressSlideshowDialog::sectorsWritten()
      * time_in_queue   milliseconds  total wait time for all requests
      */
 
+    uint numsectors=0;
+
     QFile f(sysclassblock(_drive)+"/stat");
     f.open(f.ReadOnly);
     QByteArray ioline = f.readAll().simplified();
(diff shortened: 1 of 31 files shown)


maxnet:
>A lot of the variables used to calculate disk offset, sizes and sector numbers are just integers

The MBR partition scheme also uses integers.
Bug probably just is that unsigned ones should be used.

procount:
You could be right. I didn't analyse it in detail but in some cases I also found 'int xxxx = yyy.toULongLong'.
I took a very broad brush approach ( and forgot initdrivethread.cpp ) but I needed a quick fix and it seems to work. However, a more focused expert solution is probably required before other users also report this issue.

procount:
I decided to look in a bit more detail.

NOOBS normally measures partition sizes in MiBs, so 32 bits is quite sufficient (for now anyway :D ) But for disk partitioning, this size is converted to the number of sectors (each of 512 bytes). So the maximum number of sectors that can be counted by a 32 bit signed int equates to about 1TB. Converting to unsigned can extend this only to 2TB. 

I think 3 & 4TB drives are becoming more affordable these days and users may start to use these bigger disks as storage for all their video files, considering the proliferation of media centre software for the Pi that there is these days.

maxnet:
>Converting to unsigned can extend this only to 2TB.

Correct.
That happens to be the maximum a MBR partition table can address.

Like I already mentioned they use ints as well: 

https://en.wikipedia.org/wiki/Master_boot_record#Partition_table_entries

```
+8hex		4 bytes	LBA of first absolute sector in the partition
```

Would need to implement support for GPT partition tables if you want to support anything >2TB

procount:
Thanks for the pointers and useful info @maxnet. I learn something new everyday!
OK, let's change to uints then.

procount:
```
void ProgressSlideshowDialog::setMaximum(qint64 bytes)
{
    _maxSectors = bytes/512;
    ui->progressBar->setMaximum(_maxSectors);
}
```
progressBar->setMaximum(**int**)

Should we change the scale of this progressbar to 1KB blocks instead of number of sectors, so that it can cope with >1TB ? Or just clamp it at 1TB? Or not worry?  or something else? 
Suggestions?

procount:
```
void ProgressSlideshowDialog::setMaximum(qint64 bytes)
{
    _maxSectors = bytes/512;
    ui->progressBar->setMaximum(_maxSectors);
}
```
progressBar->setMaximum(**int**)

Should we change the scale of this progressbar to 1KB blocks instead of number of sectors, so that it can cope with >1TB ? Or just clamp it at 1TB? Or not worry?  or something else? 
Suggestions?

procount:
Well, I assumed that writing 1TB of OSes is unlikely, so I just clamped it at 1TB just in case. still got no 2TB to test it on, but I'll see if I can get it verified by another means.

recovery/progressslideshowdialog.cpp
```diff
@@ -177,7 +184,9 @@ int ProgressSlideshowDialog::sectorsWritten()
     QList<QByteArray> stats = ioline.split(' ');
 
     if (stats.count() >= 6)
-        return stats.at(6).toInt(); /* write sectors */
-    else
-        return 0;
+        numsectors = stats.at(6).toUInt(); /* write sectors */
+
+    if (numsectors > 2147483647)        //Maybe use MAX_INT from limits.h?
```
XECDesign: Any reason not to use MAX_INT (for readability)?

XECDesign:
Is this tested and good to go?

procount:
I'll check if I did anymore changes to it in PINN, as it was a while ago...
(You have been busy tidying up today!)

procount:
As far as I can tell, it still looks the same in PINN. so it's tested as far as I can without a 2TB drive.
You can use MAX_INT if you prefer.

procount:
Apparently it has not been merged into NOOBS yet.
AFAICT, I didn't do any other changes to it in PINN after these PR changes

Issue #416, opened 2017-06-20
OS URL info


procount:
In each os.json, there is a "url" field. It does not seem to be used by NOOBS, but it generally points to an informative home webpage of each OS distro. So I can only assume it is there for informative reasons only, but it's not easy/obvious for a newbie user to get at this information.

I was thinking it might be nice to add a feature in NOOBS to allow the user to launch the Arora web browser at that webpage to find out more info about the OS before installing it. At the moment, this would only work with local OSes (either on SD card or USB) where the os.json file is present.

Could you copy the appropriate URLs to the os_list_v3.json file as well? Then it would also work with OSes that are downloaded, which is most of them these days. 

Thoughts?

PR #419, opened 2017-06-22
Added info button

procount:
A proposed solution to #416 which adds an Info toolbar button to jump to the webpage in the URL attribute of an OS's os.json file. This URL attribute needs to be added to the os_list_v3.json file for each OS to also work for remote repositories.


lurch:
@procount I just spotted that all the translation files in this PR say "Info (i)" when they _should_ say "Info (f)". Just letting you know in case you have the same 'bug' in PINN :wink:

procount:
Thanks, @lurch - Actually, I changed it in PINN to '(n)' to align with all my other options. I guess this won't be integrated in NOOBS now anyway, but it would need to be changed to suit NOOBS' toolbar if it were.

PR #409, opened 2017-06-03
Update Korean language by pCsOrI

bovirus:

PR #428, opened 2017-07-24
Update Korean language v2.4

sople1:
* merged #409 by my version
* correct word / text

Can I request to update?

### DIFF
--- a/recovery/translation_ko.ts
+++ b/recovery/translation_ko.ts
@@ -154,42 +154,42 @@ WARNING: This will erase all data on your SD card</source>
     <message>
         <location filename="initdrivethread.cpp" line="342"/>
         <source>Saving boot files to memory</source>
-        <translation type="unfinished"></translation>
+        <translation>부팅 파일을 메모리에 저장하고 있습니다</translation>
     </message>
     <message>
         <location filename="initdrivethread.cpp" line="345"/>
         <source>Error saving boot files to memory. SD card may be damaged.</source>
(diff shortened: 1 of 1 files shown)


recovery/translation_ko.ts
```diff
@@ -234,15 +235,15 @@ WARNING: This will erase all data on your SD card</source>
...
+        <translation>SD 카드가 손상되었습니다. (기록된 내용은 유지되지 않음)</translation>
     </message>
 </context>
 <context>
     <name>LanguageDialog</name>
     <message>
         <location filename="languagedialog.ui" line="135"/>
         <source>Language (l): </source>
-        <translation>언어 (l): </translation>
+        <translation>언어 (L): </translation>
```
XECDesign: Why is case changed here and in a few other places?

sople1: lowercase is wrong.
because Korean uses Hangul - is not Alphabet.
btw, Korean Keyboard Layout has Alphabet which is Uppercase.
https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcRORT_kXh4k0YtfBIsabgNeEs8vUPZ8CSwVbMxWgHDQktr7UzpTKQ
lowercase l - be able to confusing for I when user see keyboard.

so, it must be change to uppercase.

XECDesign: Thank you for the explanation.
I don't think I fully understand, but I will trust your judgement.

PR #438, opened 2017-09-06
Fix instructions to create custom OS

procount:
The instructions for creating a custom OS have an error when compressing the root partition. By not anchoring the exclusions in the root directory of the rootfs, it can exclude some other folders elsewhere in the filesystem tree, causing problems when the OS is later installed.

I prefixed the excluded file patterns with `./` to anchor them, as was the original intention.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ The following steps allow you to create a modified copy of one of the standard O
   2. "uncompressed_tarball_size" - replace the numerical value with the size of your filesystem tarballs when uncompressed
 
 9. Replace the `.tar.xz` root and boot filesystem tarballs with copies created from your custom OS version (these instructions assume you're only using a single OS at a time with NOOBS - they won't work if you're running multiple OSes from a single SD card). The name of these tarballs needs to match the labels given in `partitions.json`.
-  1. To create the root tarball you will need to run `tar -cvpf <label>.tar /* --exclude=proc/* --exclude=sys/* --exclude=dev/pts/*` from within the root filesystem of your custom OS version. You should then compress the resulting tarball with `xz -9 -e <label>.tar`.
+  1. To create the root tarball you will need to run `tar -cvpf <label>.tar /* --exclude=./proc/* --exclude=./sys/* --exclude=./dev/pts/*` from within the root filesystem of your custom OS version. You should then compress the resulting tarball with `xz -9 -e <label>.tar`.
   2. To create the boot tarball you will need to run `tar -cvpf <label>.tar .` at the root directory of the boot partition of your custom OS version. You should then compress the resulting tarball with `xz -9 -e <label>.tar`.
 
 ### How to change the default Language, Keyboard layout, Display mode or Boot Partition etc.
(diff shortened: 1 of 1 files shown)


XECDesign:
The command used to archive Raspbian is:
`bsdtar --numeric-owner --format gnutar --use-compress-program pxz -C ${STAGE_WORK_DIR}/rootfs --one-file-system -cpf ${NOOBS_DIR}/root.tar.xz .`
https://github.com/RPi-Distro/pi-gen/blob/dev/export-noobs/prerun.sh#L38
`--numeric-owner` makes sure permissions are preserved properly. Otherwise, groups which don't have the same ID within NOOBS and the installed OS end up being wrong.
`--use-compress-program pxz` is optional, it's just there to speed up compression, but requires pxz to be installed.
`bsdtar` and `--format gnutar` so that capabilities and long symlinks are preserved and still properly extracted by the version of tar in NOOBS.
`--one-file-system` is equivalent to those excludes, while still keeping static /dev nodes which might exist. I haven't actually checked this one, but it doesn't really matter on modern distros.

procount:
Clearly, the README is out of date and provides inaccurate instructions for users.
Should it therefore be changed to match the Raspbian archive command above, if we exclude all the pi-gen specific paths? 
Since NOOBS now uses bsdtar internally to decompress the OS, is it recommended to use bsdtar to compress it too? (Although it shouldn't really matter as they seem to be mostly compatible)

Perhaps something like the following would be sufficient with minimal impact to the doc:
`bsdtar --numeric-owner --format gnutar  --one-file-system -cpf root.tar .`
Leaving things like paths and pxz to more advanced users, who I'm sure could figure it out for themselves.

XECDesign:
I think that would be best, yes.

> Although it shouldn't really matter as they seem to be mostly compatible

Heh... 'mostly compatible'. There are different tar formats and busybox tar, gnutar and bsdtar don't handle all of them the same way. With gnutar, you end up having to pass it a page of obscure flags when creating AND extracting the archive to preserve everything that matters. Busybox tar doesn't  can't cope with some tar features and just ignored them. Then there were issues with different locales. I don't remember all the issues I've come across, but tar can be a bit of a pain.

procount:
Yes, I remember doing Arch......

XECDesign:
Just in case... has the last commit been tested?

procount:
Ha! As If I had time to test it in that time :smile: 
No, it was just a proposal for you to comment on. I removed an option for the boot partition as I didn't think it was necessary, but it wouldn't do any harm either.
I'll build a custom OS using these commands, do a test install and post back to confirm whether it is ok.
(I'll probably squash my commit as well. I know Lurch preferred clean branches)

XECDesign:
Looks alright at a glance, but I haven't tried it myself.

Github can now squash commits before merging, so there's no need to do it manually.

procount:
I converted Google's AIYProject's image to NOOBS format using the new bsdtar archive commands and it installed fine.

XECDesign:
Ship it!

Thanks

PR #443, opened 2017-10-08
Update README.md

DougieLawson:
Updated for SD card formatter version 5.0

There's no "FORMAT SIZE ADJUSTMENT" in the latest version of the formatter.


XECDesign:
Thanks

PR #452, opened 2017-11-04
Update recovery pro to add ae/no languages

bovirus:
Update recovery pro to add ae/no languages


XECDesign:
NO already present. AE doesn't exist.

Issue #456, opened 2017-11-16
LibreELEC labels


procount:
The partition labels provided in partitions.json for LibreELEC are too long. i.e. "LibreELEC_RPi_System" and "LibreELEC_RPi_Storage". 

If NOOBS finds a label longer than 15 characters, it will completely clear the label and not give the partition any label at all.

As this is an externally hosted OS, could you pass on this request to the LibreELEC maintainers to reduce their partition names to 15 characters or less?

procount:
Actually, a max length of 15 characters (allowing an extra character for an additional number to disambiguate labels with the same name) is only applicable to `ext` partitions. In `FAT` it should be 10 (+1) and in `NTFS` it is 31 (+1). Not sure about others fstypes. so the LibreELEC FAT partition should be 10 chars or less. Maybe NOOBS needs a little tweak for this?

lurch:
I did a quick search, and found the code responsible for this at https://github.com/raspberrypi/noobs/blob/master/recovery/multiimagewritethread.cpp#L452

It does seem a bit odd to completely clear the label, perhaps truncating it (again taking the destination FS into account) would be more appropriate?

BTW @procount does this actually cause install or boot problems, or is it merely a 'cosmetic' issue?

procount:
Yep, that's the spot. Sorry I should have linked to it.

AFAIK it does not cause a problem with NOOBS, except that the 2 LibreELEC partitions will have no names to identify them. So yes, it is more cosmetic than anything else.  Although if there was an OS that referenced it's partitions (e.g. in fstab) by label (which has always been a possibility in NOOBS), then that WOULD be an issue. 

I haven't checked any other OSes to see if any others are affected.

This problem was identified when testing a new feature for the next version of PINN, so it is causing me a problem and I shall have to work around it somehow. Truncating the label appropriately seems like a solution, but from which end should it be truncated in order for the remainder still to be meaningful? In the case of LibreELEC, restricting both labels to 15 characters from the start would end up with "LibreELEC_RPi2_" for both labels, so maybe truncating the beginning might be better? I don't know of any conventions here 🤷‍♂️

lurch:
One 'truncation-convention' that sometimes gets used when both the beginning and end of a filename need to be seen, is to truncate e.g. "LibreELEC_RPi_Storage" to ""LibreE...torage" - but I've got no idea if dots are allowable in filesystem labels! ;-)

XECDesign:
[...]

PR #460, opened 2017-11-23
Shorten partition label names according to fstype

procount:
If a partition label is specified that is >15 characters in length, NOOBS will unilaterally erase it.
This PR will shorten the partition label to the appropriate size for the file system instead: 11 for FAT, 16 for ext4 and 32 for NTFS. It will then make the label unique if necessary.

The shorten() function tries to make a sensible partition label out of one that has several distinct parts. If this is not required, the call can be replaced by something like `label = label.left(maxLabelLen);`

Fixes one issue from #456.


XECDesign:
Thanks

PR #462, opened 2017-11-24
Account for IOaccounting on multiDrives

### LINKED ISSUE #459: progress bar stopped working w/USB
(BUG)
Whilst testing my fixes for partition labels in a newly built NOOBS from the latest commit, I booted NOOBS from the SD card and was installing LibreELEC and another small OS from a USB stick to another USB stick on a Pi2, so the boot partitions would still be on the SD card but the rootfs would be on the USB Stick.

When it was installing them, I noticed that the progressbar on the slide show did not move - it stayed at 0% complete, and yet the installation went ahead ok. It's just that the feedback of progress was not updating.

When I removed the target USB stick and installed from USB to SD card, it worked fine.

So I downloaded the latest NOOBS_lite 2.4 from the web and tried with that to test if my changes had caused the problem but the same thing happened. I will investigate a little further whilst testing my fixes and report if I find the cause.

procount:
The progressSlideShowDialog shows the progress of installation by monitoring the number of sectors written to the installation _drive. However, now that an OS rootfs can be written to a USB drive, any sectors written to the _bootDrive are not accounted for. This is very noticeable for an OS like LibreELEC where ALL of the image is written to the boot partition and the rootfs is purely used for storage, so the progress bar hardly moves during installation.

This PR fixes this by allowing the drive that is being monitored to be changed during the installation of the various partitions.

Fixes #459


XECDesign:
Thanks

PR #469, opened 2018-01-31
Change wpa_supplicant copy

### LINKED ISSUE #341: NOOBS wifi preconfig
In commit  f0347db8d0eece99e97b97ab8ed32a7d584ea0c4 "Add Wfif Station Support" (oct 26 2015) a great new feature was added to copy the wpa_supplicant.conf file from the NOOBS partition to the /settings partition which allows the wifi settings for NOOBS to be preconfigured. This feature has never really been announced, although the similar feature to copy the file from the /boot partition was announced for the latest Raspbian, 
Even though I documented this feature in my update to 1.9, and announced it on the forum, there are still some people tweeting and posting that are not aware that this feature is available in NOOBS. So I think it should be announced better. **BUT BEFORE YOU DO....** there are a couple of improvements that need making.
1. In mainwindow.cpp#911 (startNetworking), the `if ( !f.exists() )` statement means that any user's wifi config file is ONLY ever copied the first time NOOBS is booted. I suspect this is to avoid copying `/etc/wpa_supplicant.conf` over the top of a user configuration on a subsequent invocation. But it prevents a user from _**ever**_ modifying the wifi config to be used by NOOBS without first deleting `/settings/wpa_supplicant.conf` (not easy for a Newbie!). So I think the logic is a bit wrong. Better to ALWAYS copy a users wifi config to /settings, and only copy the default one if it doesn't already exist 
   in /settings. (Maybe even delete or rename the user's config file after copying it to avoid overwriting any manual wifi entries that are later added in the gui) That way the user can update the preconfiguration.
2. When Raspbian is installed from NOOBS, it should also look at the /settings partition for wifi configuration and install that file like it does from the /boot partition. This can be done with a simple change to `/os/Raspbian/partition_setup.sh`, as detailed in my [forum post](https://www.raspberrypi.org/forums/viewtopic.php?f=63&t=141559)

procount:
To fix #341 part 1.
Diff makes this look more complicated than it is.
Essentially the copy of wpa_supplicant.conf has been taken out of startNetwokring() into copywpa() which is then called ahead of startNetworking() regardless of silentinstall option.
Then copywpa() was modified to provide the necessary functionality.

(This PR is a copy/paste out of PINN into NOOBS)

### DIFF
--- a/recovery/mainwindow.cpp
+++ b/recovery/mainwindow.cpp
@@ -215,6 +215,9 @@ MainWindow::MainWindow(const QString &drive, const QString &defaultDisplay, QSpl
     {
         _fixate = true;
     }
+
+    copywpa();
+
     if (cmdline.contains("silentinstall"))
     {
         /* If silentinstall is specified, auto-install single image in /os */
@@ -951,29 +954,48 @@ void MainWindow::on_list_doubleClicked(const QModelIndex &index)
     }
 }
 
-void MainWindow::startNetworking()
+void MainWindow::copyWpa()
 {
+    //This file is the one used by dhcpcd
     QFile f("/settings/wpa_supplicant.conf");
-
     if ( f.exists() && f.size() == 0 )
     {
         /* Remove corrupt file */
         f.remove();
     }
-    if ( !f.exists() )
+
+    /* If user supplied a wpa_supplicant.conf on the FAT partition copy that one to settings regardless */
+    if (QFile::exists("/mnt/wpa_supplicant.conf"))
     {
-        /* If user supplied a wpa_supplicant.conf on the FAT partition copy that one to settings
-           otherwise copy the default one stored in the initramfs */
-        if (QFile::exists("/mnt/wpa_supplicant.conf"))
-            QFile::copy("/mnt/wpa_supplicant.conf", "/settings/wpa_supplicant.conf");
-        else
-        {
-            qDebug() << "Copying /etc/wpa_supplicant.conf to /settings/wpa_supplicant.conf";
-            QFile::copy("/etc/wpa_supplicant.conf", "/settings/wpa_supplicant.conf");
-        }
+        qDebug() << "Copying  user wpa_supplicant.conf to /settings/wpa_supplicant.conf";
+
+        QProcess::execute("mount -o remount,rw /settings");
+        QProcess::execute("mount -o remount,rw /mnt");
+
+        QFile::remove("/settings/wpa_supplicant.conf.bak");
+        QFile::rename("/settings/wpa_supplicant.conf","/settings/wpa_supplicant.conf.bak");
+        QFile::copy("/mnt/wpa_supplicant.conf", "/settings/wpa_supplicant.conf");
+        f.setPermissions( QFile::WriteUser | QFile::ReadGroup | QFile::ReadOther | QFile::ReadUser );
+
+        /* rename the user file to avoid overwriting any manually set SSIDs */
+        QFile::remove("/mnt/wpa_supplicant.conf.bak");
+        QFile::rename("/mnt/wpa_supplicant.conf","/mnt/wpa_supplicant.conf.bak");
+
+        QProcess::execute("sync");
+        QProcess::execute("mount -o remount,ro /settings");
+        QProcess::execute("mount -o remount,ro /mnt");
+    }
+    else if ( !f.exists() )
+    {
+        /* There is no existing file, must be first installation */
+        qDebug() << "Copying /etc/wpa_supplicant.conf to /settings/wpa_supplicant.conf";
+        QFile::copy("/etc/wpa_supplicant.conf", "/settings/wpa_supplicant.conf");
     }
     QFile::remove("/etc/wpa_supplicant.conf");
+}
 
+void MainWindow::startNetworking()
+{
     /* Enable dbus so that we can use it to talk to wpa_supplicant later */
     qDebug() << "Starting dbus";
     QProcess::execute("/etc/init.d/S30dbus start");
--- a/recovery/mainwindow.cpp
+++ b/recovery/mainwindow.cpp
@@ -973,13 +973,13 @@ void MainWindow::copyWpa()
         QProcess::execute("mount -o remount,rw /settings");
         QProcess::execute("mount -o remount,rw /mnt");
 
-        backup("/settings/wpa_supplicant.conf");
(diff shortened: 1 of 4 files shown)


recovery/mainwindow.cpp
```diff
@@ -215,6 +215,9 @@ MainWindow::MainWindow(const QString &drive, const QString &defaultDisplay, QSpl
     {
         _fixate = true;
     }
+
+    copywpa();
```
lurch: Should this have a capital `W` ?

recovery/mainwindow.cpp
```diff
@@ -951,29 +954,48 @@ void MainWindow::on_list_doubleClicked(const QModelIndex &index)
...
-        /* If user supplied a wpa_supplicant.conf on the FAT partition copy that one to settings
-           otherwise copy the default one stored in the initramfs */
-        if (QFile::exists("/mnt/wpa_supplicant.conf"))
-            QFile::copy("/mnt/wpa_supplicant.conf", "/settings/wpa_supplicant.conf");
-        else
-        {
-            qDebug() << "Copying /etc/wpa_supplicant.conf to /settings/wpa_supplicant.conf";
-            QFile::copy("/etc/wpa_supplicant.conf", "/settings/wpa_supplicant.conf");
-        }
+        qDebug() << "Copying  user wpa_supplicant.conf to /settings/wpa_supplicant.conf";
```
lurch: Two spaces between "Copying" and "user" ;)

recovery/mainwindow.cpp
```diff
@@ -951,29 +954,48 @@ void MainWindow::on_list_doubleClicked(const QModelIndex &index)
...
+
+        QProcess::execute("mount -o remount,rw /settings");
+        QProcess::execute("mount -o remount,rw /mnt");
+
+        QFile::remove("/settings/wpa_supplicant.conf.bak");
```
lurch: Perhaps it'd be worth making a little `backup` utility function which does something like (pseudocode):
```
bool backup(filename) {
    backupname = filename + '.bak';
    if (exists(backupname)) {
        remove(backupname);
    }
    return rename(filename, backupname);
}
```

recovery/mainwindow.cpp
```diff
@@ -951,29 +954,48 @@ void MainWindow::on_list_doubleClicked(const QModelIndex &index)
...
+        QFile::rename("/settings/wpa_supplicant.conf","/settings/wpa_supplicant.conf.bak");
+        QFile::copy("/mnt/wpa_supplicant.conf", "/settings/wpa_supplicant.conf");
+        f.setPermissions( QFile::WriteUser | QFile::ReadGroup | QFile::ReadOther | QFile::ReadUser );
+
+        /* rename the user file to avoid overwriting any manually set SSIDs */
```
lurch: IMHO this comment is a bit confusing - is it essentially the same as saying "rename the user file to indicate that it has been copied (and prevent it being re-copied next time)" ?

Although I guess this breaks the "NOOBS prime directive" of not modifying any files on the RECOVERY partition. Although I've not been keeping up to date with NOOBS for a while, so I dunno if that prime-directive is still in place? ;)

recovery/mainwindow.cpp
procount: Yep. Don't know where it went - it's there in my original code. Odd.

recovery/mainwindow.cpp
procount: Eagle-eyed! Yes.

recovery/mainwindow.cpp
procount: maybe...

recovery/mainwindow.cpp
procount: Yes. I was concentrating more on the consequences of it being copied a second time.

Dunno either, but I wasn't too concerned about that in PINN. Not sure if there's another way around it.

procount:
Any better? I've compiled it on another branch, but not tested.

recovery/util.cpp
```diff
@@ -39,6 +39,17 @@ void putFileContents(const QString &filename, const QByteArray &data)
     f.close();
 }
 
+bool backup(QString filename, QString ext)
+{
+
+    QString backupName = filename + "." + ext;
+    const char * backupFile = backupName.toUtf8().constData();
```
lurch: Just out of curiosity, why the switch away from the Qt functions (QString, QFile::remove, QFile::rename, etc.) ? I think that _might_ allow you to get rid of the ugly `.toUtf8().constData()` ?

recovery/util.h
```diff
@@ -16,6 +16,7 @@
 
 QByteArray getFileContents(const QString &filename);
 void putFileContents(const QString &filename, const QByteArray &data);
+bool backup(QString filename, QString ext="bak");
```
lurch: Perhaps this should be `const QString &filename` just like `putFileContents` uses? ;-)

recovery/mainwindow.cpp
lurch: I guess with having the same configuration file stored / copied / edited in two different places, there'll never be a "perfect" way of doing things, so it's probably one of those "pick the least worst" situations ;)

recovery/util.cpp
procount: Sorry, I took your pseudo code too literally :) - I will revert.

recovery/util.cpp
```diff
@@ -39,15 +39,15 @@ void putFileContents(const QString &filename, const QByteArray &data)
...
-    const char * backupFile = backupName.toUtf8().constData();
-    if( access( backupFile, F_OK ) != -1 ) {
-        remove(backupFile);
+
+    if (QFile::exists(backupName))
+    {
+        QFile::remove(backupName);
     }
-    return rename(filename.toUtf8().constData(), backupFile);
+    return  QFile::rename(filename,backupName);
```
lurch: Woohoo much neater :tada:

recovery/util.h
```diff
@@ -16,7 +16,7 @@
 
 QByteArray getFileContents(const QString &filename);
 void putFileContents(const QString &filename, const QByteArray &data);
-bool backup(QString filename, QString ext="bak");
+bool backup(const QString &filename, const QString &ext="bak");
```
lurch: Looking at the other functions declared here, perhaps `backupFile` would be a better name? :)
(sorry for being so nit-picky!)

XECDesign:
Thanks

PR #474, opened 2018-02-15
Update rpi-wifi-firmware

### LINKED ISSUE #473: rpi-wifi-firmware package issues
@XECDesign : There are some issues with the rpi-wifi-firmware package (especially for new developers cloning the noobs github and expecting it to build).

1. rpi-wifi-firmware.mk uses 'master' as the version number, so it will always pull down the latest version rather than a specific version. **(Fix: use a specific commit)**

2. The repo this points to has been upstreamed(*) and the folders have been reorganised. This means it is no longer compatible with rpi-wifi-firmware.mk and it no longer "builds" properly. **(fix: update rpi-wifi-firmware.mk to the new folder structure)**

3. brcmfmac43430-sdio.txt now exists as a patch (https://github.com/RPi-Distro/firmware-nonfree/commit/8d4f3825be09e556c01543119ed50515bd7dd661#comments), so it does not get included by default in the build **(fix: ???)**

So I can resolve points 1&2 as indicated above, but how to fix point 3 so that NOOBS builds "out of the box" again? It is possible to copy the patch into buildroot/package/rpi-wifi-firmware and add it to the NOOBS github. This works, but it won't get updated with the Rpi-Distro repo version.

Is it possible to automate the copying or updating of this patch somehow? It seems a bit of a catch-22 scenario to me: The patch isn't present until the rpi-wifi-firmware is downloaded/patched/configured/built etc by buildroot, but it needs to patch the package before it is configured & built. Somehow the copy needs to take place between downloading the package and patching the package to avoid 2 passes of the build. Any ideas?

(*) - maybe the wrong terminology?

XECDesign:
Fixes #473

### DIFF
--- a/buildroot/package/rpi-wifi-firmware/rpi-wifi-firmware.mk
+++ b/buildroot/package/rpi-wifi-firmware/rpi-wifi-firmware.mk
@@ -4,15 +4,15 @@
 #
 ################################################################################
 
-RPI_WIFI_FIRMWARE_VERSION = master
+RPI_WIFI_FIRMWARE_VERSION = 927fa8ebdf5bcfb90944465b40ec4981e01d6015
 RPI_WIFI_FIRMWARE_SITE = $(call github,RPi-Distro,firmware-nonfree,$(RPI_WIFI_FIRMWARE_VERSION))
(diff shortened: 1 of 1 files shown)


XECDesign:
Thanks for checking.

PR #477, opened 2018-02-25
Drop dependencies of qt-linguist tool from host

mcapdeville:
Recovery need lrelease and lupdate from host for updating translation.
This is done in recovery/updateqm.sh . 

The first patch add the option BR2_PACKAGE_QT_HOST_TOOLS to buildroot for building the Qt tools target that contain the qt-linguist tool, so lrelease and lupdate are make available in ${HOST_DIR}/usr/bin .

The second patch add the BR2_PACKAGE_QT_HOST_TOOLS build dependeciy to the recovery package.

The third patch modify updateqm.sh to get lrelease and lupdate from buildroot ${HOST_DIR} .


lurch:
See also #158 ;-)

XECDesign:
Then I think the pre-commit hook would also need to be updated to use lupdate from buildroot if it's built or the system version otherwise.

XECDesign:
Inactive

PR #496, opened 2018-05-04
Port of self Update from PINN

procount:
Just in case there's a change of heart, I've ported my self update code from PINN to NOOBS.

The 2 constants README_URL and BUILD_URL in config.h need to be modified to wherever you wish to locate the README.md and BULD-DATA files in your CDN system. Currently they point at my github for testing. 

It only updates the NOOBS files, not any OSes in /os. That feature would need adding if you want it.

There is a `no_update` option in recovery.cmdline to disable self updating.

### DIFF
--- a/recovery/builddata.cpp
+++ b/recovery/builddata.cpp
@@ -1,0 +1,170 @@
+#include "builddata.h"
+#include <QProcess>
+#include <QRegExp>
+#include <QFile>
+#include <QDebug>
+
+BuildData::BuildData()
+{
+}
+
+void BuildData::splitVersion(void)
+{
+    QString::iterator it = _version.begin();
+    while (it != _version.end())
+    {
+        int digitmode = it->isDigit();
+        QString part="";
+
+        if (digitmode)
+            while (it != _version.end() && (*it).isDigit())
+                part += *it++;
+        else
+            while (it != _version.end() && !(*it).isDigit())
+                part += *it++;
+        _partslist <<part;
+        digitmode = !digitmode;
+    }
+}
+
+void BuildData::setVersion(const char * v)
+{
+    _version =v;
+    splitVersion();
+}
+
+bool BuildData::isEmpty(void)
+{
+    bool result = _version.isEmpty();
+    return (result);
+}
+
+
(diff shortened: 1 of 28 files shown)


recovery/builddata.h
```diff
@@ -0,0 +1,28 @@
...
+#include <QString>
+#include <QStringList>
+
+class BuildData
+{
+public:
+    BuildData();
+    void read(const char * filename);
+    bool isEmpty(void);
+    void getVersionParts(void);
```
lurch: Seems like this function isn't implemented??

recovery/builddata.cpp
```diff
@@ -0,0 +1,170 @@
...
+#include <QProcess>
+#include <QRegExp>
+#include <QFile>
+#include <QDebug>
+
+BuildData::BuildData()
+{
+}
+
+void BuildData::splitVersion(void)
```
lurch: Probably makes sense to just move this functionality wholesale to inside `setVersion` ?

recovery/builddata.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+        if (digitmode)
+            while (it != _version.end() && (*it).isDigit())
+                part += *it++;
+        else
+            while (it != _version.end() && !(*it).isDigit())
+                part += *it++;
+        _partslist <<part;
+        digitmode = !digitmode;
+    }
```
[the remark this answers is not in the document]
procount: Redundant remnants of an earlier incantation

recovery/builddata.h
procount: forgot to remove

recovery/builddata.cpp
procount: I think I previously called splitVersion() from elsewhere, but didn't refactor it out in later iterations

procount:
@lurch -thanks for reviewing, but I'm not sure if it is worth while fixing any changes here if it's not going to be implemented. I have removed the old, redundant code from PINN's next version though to clean it up, so much appreciated! 👍

XECDesign:
Sorry, closing because it goes outside of the cope of NOOBS.

PR #497, opened 2018-05-07
Use partuuid in installed_os.json

procount:
When installing an OS to a USB storage device, the partitions used for the OS are referenced assuming the device remains consistent as /dev/sdaX. I had one user who booted PINN with 2 USB sticks - one with the OS installations on it (sda), and the other used as the destination (sdb). Naturally the partition was stored as /dev/sdb6 in installed_os.json, but on later boots, the device was referenced as /dev/sda so the OS could not be found.

I think that OSes installed to an external USB should be referenced by partuuid in os_config.json and installed_os.json, just like is done in cmdline.txt and /etc/fstab.

Do you think the attached Pull Request will solve this issue? Are there any problems with it?


procount:
In order for the 1st commit to compile, I moved 4 uuid related functions from multiimagewritethread.cpp to util.cpp.

recovery/multiimagewritethread.cpp
```diff
@@ -530,7 +530,13 @@ bool MultiImageWriteThread::processImage(OsInfo *image)
     QVariantList vpartitions;
     foreach (PartitionInfo *p, *partitions)
     {
-        vpartitions.append(p->partitionDevice());
+        QString part = p->partitionDevice();
+        if (part.left(13) != "/dev/mmcblk0p")
```
lurch: Just out of curiosity, what happens if you remove this special-casing for `/dev/mmcblk0` and instead _always_ use the partuuid?

lurch:
> In order for the 1st commit to compile, I also had to...

Might be beneficial to swap the order of the commits?

recovery/multiimagewritethread.cpp
maxnet: We used to have some Linux distributions that came with an initramfs that did not support partuuid parameters (e.g. OSMC).
Not sure if that's still the case in recent versions though.

procount:
If I were to rebase to reorder the commits (or even squash them), wouldn't it wreck all this conversation, comments, history etc when I push it up (probably with -f to force it?)

lurch:
Nope, I don't think so.

recovery/multiimagewritethread.cpp
lurch: @procount I wonder if it'd be worth adding an extra param to the `os.json` / `os_list.json` files stipulating whether they support partuuid parameters, and then only using this partuuid code if they do?

procount:
OK I'm willing to give it a go - Which do you prefer: swap or squash?

recovery/multiimagewritethread.cpp
procount: Can we piggy-back the `usb_root` or `usb_boot` parameters for that? I think they serve a similar purpose...?

procount: IIRC, if an OS doesn't support partuuid, then it can't be installed to USB. If so, then the additional code won't be necessary. I'm sure @XECDesign  explained the use of `supports_usb_root` and `supports_usb_boot` somewhere but I can't find the reference.

procount:
I swapped first, then squashed, and both worked, so I guess the git warnings about rebasing and force pushing only apply to the public repository that others have used and not to pull requests.

XECDesign:
@maxnet, does this look okay to you?

maxnet:
Looks ok, although I haven't tried it.

XECDesign:
Thanks for taking a look. I'll test it out before it goes in a release.

recovery/bootselectiondialog.cpp
```diff
@@ -157,13 +157,33 @@ void BootSelectionDialog::accept()
     QSettings settings("/settings/noobs.conf", QSettings::IniFormat, this);
     QVariantMap m = item->data(Qt::UserRole).toMap();
     QByteArray partition = m.value("partitions").toList().first().toByteArray();
-    QRegExp partnrRx("([0-9]+)$");
-    if (partnrRx.indexIn(partition) == -1)
+
+    int partitionNr;
+    QRegExp parttype("^PARTUUID");
+    if (parttype.indexIn(partition) == -1)
```
lurch: You could probably use http://doc.qt.io/archives/qt-4.8/qstring.html#startsWith here instead?

procount:
@XECDesign - I finally got around to testing this, with boot on SD card and root on USB, but it didn't use partuuids, so I'll try & find out why...

procount:
@XECDesign - I finally got around to testing this with boot on sd card and root on USB stick. 
It correctly used PARTUUIDS for the USB partitions and interpreted the HEX partition numbers correctly.
(Of course, it helps if you build and test from the correct GIT branch!)

PR #503, opened 2018-06-05
Automatic keyboard layout change for Swedish

Mangero:
Added if statement to languagedialog.cpp in the same way as for
Japanese to automatically switch to Swedish keyboard layout
when Swedish language is chosen.

This is the same change as in commit f3b6970 of pull request #252

### DIFF
--- a/recovery/languagedialog.cpp
+++ b/recovery/languagedialog.cpp
@@ -182,6 +182,10 @@ void LanguageDialog::changeLanguage(const QString &langcode)
     {
         defaultKeyboardLayout = "jp";
     }
+    else if (langcode == "sv")
+    {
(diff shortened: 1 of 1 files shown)


XECDesign:
@Mangero This has been tested and works as expected?

Mangero:
@XECDesign 
I did test this thoroughly before I submitted my first pull request, that was a few years ago. I have not tested it with a current build but since this is just the addition of an if-statement which works the exact same way as the ones already in place for dutch and japanese I can't imagine there could be any problems. If you'd like me to test it with the current build I might be able to do that some time during the weekend.

XECDesign:
Thank you

PR #523, opened 2018-12-08
Update README.md

Raymo111:
Apt is newer and more user-friendly than apt-get

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,7 @@ Note that this will require a minimum of 6GB free disk space.
 
 On Ubuntu:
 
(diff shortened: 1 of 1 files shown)


lurch:
@XECDesign Probably worth merging this, given that we've recently updated the documentation repo to switch from `apt-get` to `apt` ?

XECDesign:
Definitely, thanks for the nudge. Can't merge right now though, since there are a ton of local uncommitted changes.

Raymo111:
@XECDesign Keep us updated!

PR #545, opened 2019-06-11
Updated the GPIO Trigger functionality to allow custom GPIO ports

mjjames:
You can now provide the GPIO number and the value you want it to be (e.g 1 or 0) allowing more use case scenarios


XECDesign:
In general, we don't accept PRs which add functionality, but this looks simple enough.

If you also add some documentation for the feature, remove the NOOBS version bump and  it all works for you as expected, I'm happy to merge. 

Thanks.

mjjames:
Reverted to V3 and added additional documentation underneath the existing GPIO Trigger part. Added examples so that its hopefully clear

mjjames:
abandoning the PR as I pushed other changes to master, instead created a new PR which has a specific branch containing the changes #546

PR #548, opened 2019-07-03
README.md: Add instructions on enabling headless SSH

ProactiveServices:


XECDesign:
Doesn't just dropping an ssh file onto the recovery partition work?

ProactiveServices:
On the pre-installed NOOBS SD card that I received, no. I presumed that the pre-installation environment didn't support it as it wasn't a full operating system. Once I'd made the changes to have NOOBS auto-install the OS, the ssh file successfully caused the system to start the SSHd.

procount:
I'm pretty sure dropping the ssh file into the recovery partition worked last time I installed Raspbian with PINN, post-pi4, so it should have worked in NOOBS too.

procount:
Ah wait a minute, was Raspbian already PRE-INSTALLED? I'm which case it wouldn't work unless you reinstall Raspbian cos it only copies the file on reinstallation in the partition setup script.

ProactiveServices:
Sorry for any confusion. Raspbian wasn't pre-installed, NOOBS was. I have since made changes to start using the Pi but will be able to test this as I have an image of the original data.

XECDesign:
Neither the steps in this PR or dropping the ssh file would help if Raspbian isn't installed afterwards.

Given that partitions_setup.sh already has a way of handling ssh and wpa_supplicant.conf files, I wouldn't recommend modifying it.

https://github.com/RPi-Distro/pi-gen/blob/master/export-noobs/00-release/files/partition_setup.sh#L23

PR #560, opened 2019-11-08
Update dependencies of recovery.mk

procount:
recovery.mk modifies the file target/libexec/dhcpcd-hooks/10-wpa-supplicant generated by the dhcpcd package.
By chance the dhcpcd package is normally built before recovery, but minor alterations to other dependencies can cause the build order to change. This change makes the dependency explicit to avoid future build errors.


XECDesign:
Thanks. I'll merge this when doing the next release.

RaspberryPiFan:
I'll add that to my PR on PINN for buildroot update

PR #561, opened 2019-11-25
Create LICENSE

Bobbbay:
Yes, I am aware that NOOBS already has a license.txt file in /recovery/LICENSE.txt, but this license is not recognised by Github because of the odd path and therefore Github thinks that NOOBS doesn't have an official license.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,15 @@
+Copyright (c) 2013, Raspberry Pi
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
+
+Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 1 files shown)


XECDesign:
That license only applies to the recovery directory, so moving it outside doesn't make sense, regardless of how github work.

Bobbbay:
So then should there be an official license on the files outside of the recovery module (e.x. MIT or Apache)? I think there should be a license on one of the world's biggest bootloaders... Unless, of course a missing license was done deliberately?

XECDesign:
buildroot comes with its own COPYING file and each package provided by buildroot should have its own license information in the makefile.

Bobbbay:
Understandable, so every file has its own license. Got it.

lurch:
More info at https://github.com/raspberrypi/noobs#legal-compliance and https://github.com/raspberrypi/noobs#third-party-licenses

PR #562, opened 2019-11-25
Create CODE_OF_CONDUCT.md

Bobbbay:
A code of conduct is recommended for the transperency of any project. This is a quick add and can be edited if needed, but this is the template Github gives, so I would personally recommend this.


XECDesign:
Thank you, but we don't currently have a standard code of conduct for our projects. If the powers that be deem it necessary, I'm sure one will be added.

PR #563, opened 2019-12-05
Add option update to update wifi-firmware to BUILDME.sh

RaspberryPiFan:
This makes it easier to update rpi-eifi-firmware and adds it to build data.

### DIFF
--- a/BUILDME.sh
+++ b/BUILDME.sh
@@ -127,9 +127,9 @@ for i in $*; do
         update_github_kernel_version raspberrypi/linux rpi-4.14.y
     fi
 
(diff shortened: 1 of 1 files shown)


lurch:
I'm not sure what LibreELEC has to do with anything??

Looking at  https://github.com/raspberrypi/noobs/blob/master/buildroot/package/rpi-wifi-firmware/rpi-wifi-firmware.mk would a better approach be to try and get newer firmware into https://github.com/RPi-Distro/firmware-nonfree ?

RaspberryPiFan:
Yeah, I looked at the default buildroot, which uses LibrELEC

RaspberryPiFan:
Fixed.

PR #564, opened 2019-12-09
Less restrictive search for NOOBS files

### LINKED ISSUE #498: BUG: "Cannot find the drive with NOOBS files"
See also #467 
I think I figured out why this error is occurring.
IF a user formats their SD card as straight FAT with no MBR, then it is mounted as /dev/mmcblk0.
The check in findRecoveryDrive() to find the recovery partition only looks at partition 1 of each device (i.e. /dev/mmcblk0p1) which does not exist, and therefore the recovery files cannot be found.

The function InitDriveThread::method_resizePartitions() can detect a missing mbr and reformat the SD card appropriately, but requires the user to re-copy the NOOBS files back onto it afterwards. In any case, this is run AFTER findRecoveryDrive().

procount:
Fixes #498 

This removes some restrictions on the drives that NOOBS will search for its files to determine the boot drive.
In this way it will detect NOOBS on a superfloppy drive that has no MBR and will allow it to be partitioned and formatted properly.

I produced some test images to prove it works.

https://sourceforge.net/projects/pinn/files/testing/superfloppy.zip/download is an image file that should be flashed to a blank SD card with Balena Etcher. It contains NOOBS 3.0 on a superfloppy formatted partition. When it boots, it will not detect the NOOBS files.

https://sourceforge.net/projects/pinn/files/testing/superfloppyfix.zip/download is an image file that should be flashed to a blank SD card with Balena Etcher. It is the same as the previous image, except I replaced recovery.rfs with a version of NOOBS recovery program with this pull request in. It shows that the drive will be detected, partitioned and formatted correctly.

### DIFF
--- a/recovery/main.cpp
+++ b/recovery/main.cpp
@@ -81,9 +81,8 @@ QString findRecoveryDrive()
 
     foreach (QString devname, list)
     {
-        /* Only search first partition and partitionless devices. Skip virtual devices (such as ramdisk) */
-        if ((devname.right(1).at(0).isDigit() && !devname.endsWith("1"))
-                || QFile::symLinkTarget("/sys/class/block/"+devname).contains("/devices/virtual/"))
+        /* Skip virtual devices (such as ramdisk) */
+        if (QFile::symLinkTarget("/sys/class/block/"+devname).contains("/devices/virtual/"))
             continue;
 
         if (QProcess::execute("mount -t vfat -o ro /dev/"+devname+" /mnt") == 0)
--- a/recovery/main.cpp
+++ b/recovery/main.cpp
@@ -81,9 +81,31 @@ QString findRecoveryDrive()
 
     foreach (QString devname, list)
     {
-        /* Skip virtual devices (such as ramdisk) */
-        if (QFile::symLinkTarget("/sys/class/block/"+devname).contains("/devices/virtual/"))
+        if (QFile::symLinkTarget("/sys/class/block/"+devname).contains("/devices/virtual/") )
+        {   //Skip if partition is virtual.
             continue;
+        }
+
+        int partitionNr = 0;
+        QRegExp partnrRx("([0-9]+)$");
+        if (partnrRx.indexIn(devname) != -1)
+        {   //Does devname end in 1 or more digits? Yes->Read it
+            partitionNr = partnrRx.cap(1).toInt();
+
+            if (partitionNr>1)
+            {   //Skip if partition Number >1
+                continue;
+            }
+        }
+
+        if (partitionNr == 0)
+        {   //Check for a possible non-MBR drive
+            if ( list.contains(devname+"1") || list.contains(devname+"p1") )
+            {
+                //skip if partition 1 exists (must have MBR)
+                continue;
+            }
+        }
 
         if (QProcess::execute("mount -t vfat -o ro /dev/"+devname+" /mnt") == 0)
         {
@@ -106,10 +128,8 @@ QString findRecoveryDrive()
         if (!drive.isEmpty())
             break;
     }
-
-    return drive;
+    return(drive);
 }
-
 
 int main(int argc, char *argv[])
 {


recovery/main.cpp
```diff
@@ -81,9 +81,8 @@ QString findRecoveryDrive()
 
     foreach (QString devname, list)
     {
-        /* Only search first partition and partitionless devices. Skip virtual devices (such as ramdisk) */
-        if ((devname.right(1).at(0).isDigit() && !devname.endsWith("1"))
```
lurch: Ahhhh, now I finally see why the existing "partitionless devices" check was failing! :laughing: 
It was checking that the devname either doesn't end with a digit (e.g. `sda` for a partitionless USB drive), or it ends with 1 (e.g. `sda1` for a partitioned USB drive). This ensures that it doesn't erroneously "find" `recovery.rfs` e.g. on `/dev/sda3`.

But when we're booting from an SD-card, both  the partitionless device (i.e. `mmcblk0`) **and** the partitioned device (i.e. `mmcblk0p1`) both end in a digit!

So @procount I wonder if a better fix would be to change it to
```
if ((devname.right(1).at(0).isDigit() && !(devname.endsWith("1") || devname.endsWith("0")))
    || QFile::symLinkTarget("/sys/class/block/"+devname).contains("/devices/virtual/"))
```
:question: 

(I'm afraid I don't have tine to test this myself)

procount:
I believe the latest bootloaders don't stop at partition 1, but will continue searching until it finds bootcode.bin on any FAT partition, so why stop at 1?

procount:
@lurch if you want to add this constraint, your expression currently doesn't guard against mmcblk0p10,11,20,21 etc.

lurch:
> your expression currently doesn't guard against mmcblk0p10,11,20,21 etc.

Yeah I thought of that too, but the original expression didn't guard against mmcblk0p11 or mmcblk0p21 either, and I guess 10 or more partitions is quite unusual? :wink: 

> Except for the fact that noobs only lives on part1 I guess.

Yep.

> It might save some scan time, but will it make much difference?

Dunno, I'm not sure why that check was added in the first place. Maybe it guards against the (probably unlikely) possibility that one of the OSes installed _by_ NOOBS has a file named `recovery.rfs` on its FAT partition? :man_shrugging:  And as you say, reducing the scan time doesn't sound like a bad thing?

procount:
```
        int partitionNr = 99;
        QRegExp partnrRx("([0-9]+)$");
        if (partnrRx.indexIn(devname) != -1)
        {
            partitionNr    = partnrRx.cap(1).toInt();
        }

        if ( (partitionNr>1) ||
             (QFile::symLinkTarget("/sys/class/block/"+devname).contains("/devices/virtual/") )
            continue;
```

recovery/main.cpp
```diff
@@ -81,10 +81,31 @@ QString findRecoveryDrive()
...
+        {   //Does devname end in 1 or more digits? Yes->Read it
+            partitionNr = partnrRx.cap(1).toInt();
+
+            if (partitionNr>1)
+            {   //Skip if partition Number >1
+                continue;
+            }
+        }
+
+        if (partitionNr == 0)
```
lurch: I guess technically you could have `int partitionNr = -1;` on line 89, and then you could do `if ( ((partitionNr == -1) && list.contains(devname+"1")) || ((partitionNr == 0) && list.contains(devname+"p1")) )` (or spilt it into multiple `if`s), but maybe that's an optimisation that it's not worth making? :man_shrugging:

recovery/main.cpp
```diff
@@ -107,11 +128,9 @@ QString findRecoveryDrive()
         if (!drive.isEmpty())
             break;
     }
-
-    return drive;
+    return(drive);
```
lurch: Why this change? :wink:

recovery/main.cpp
procount: Yes, you could, but I don't think it's worth it.
When I tested on Ubuntu, it also found /dev/sr0 which got scanned, but I'm happy with it as is.

recovery/main.cpp
procount: Yes, I spotted that too. (Cut and paste from rewritten test program).
I wondered if you would comment..... 👀 😉

PR #568, opened 2020-01-17
Galician translation

alexandregz:
Galician language (ISO code gl)


XECDesign:
Thank you. Will merge before next NOOBS release.

XECDesign:
Looks like we forgot to add it to recovery.pro, so it didn't quite make it into NOOBS 3.3.

lurch:
Looking at https://github.com/raspberrypi/noobs#addingupdating-translations I guess `icons.qrc` should have been updated too.

procount:
Is Galician another Spanish dialect?
I think GL the ISO country code for greenland.....

XECDesign:
Yup, you're right.

XECDesign:
I'll need to read qt's localisation documentation again later, since I've forgotten a lot of how it, but at a glance, it looks like  `gl` may be right after all. We're specifying the language, not the country. The language is a ISO 639 code, which is indeed `gl` for Galician. `GL` is the ISO 3166-1 alpha-2 country code for Greenland. Qt seems to take `language[_script][_country][.codeset][@modifier]` in the QLocale constructor. So, I _think_ we're okay, but will have to double check when I have time.

lurch:
https://en.wikipedia.org/wiki/Galician_language confirms that `gl` is the ISO 639-1 language code for Galician. Note that language-codes and country-codes are different, and frequently overlap and/or clash :upside_down_face:

XECDesign:
I bet we have the same set of wikipedia tabs open.

procount:
Yep, it's been a while since I last looked too, so it's probably ok.

alexandregz:
Hi, galician is a language, not a dialect 

It's very close to Portuguese, in fact Galician is the origin of Portuguese 😄

PR #576, opened 2020-03-22
Update translation to russian

Morganov:
- Fix mistakes
  - Add strings translation
  - Improve existing translation


XECDesign:
That reads much better, thanks.

Issue #577, opened 2020-03-23
Shall README 'Setup' mention that Pi 4B may get video only in one HDMI port?


aszasz:
Once the Setup section of the README.md is thoughtful enough to say "Please note that in some cases it may extract the files into a folder, if this is the case then please copy across the files from inside the folder rather than the folder itself."

I think it would be equally worth to mention that "For Pi 4 Model B, it has been reported that switching to HDMI port closer to the power supply has solved the lack of video response"

That is because it took me long to try the swicth, as well as other reported here ( https://raspberrypi.stackexchange.com/questions/101842/noobs-stuck-at-rainbow-screen-raspberry-pi-4 )

(I also requested raspberrypi.org to mention it on the set-up guide)

lurch:
A side-effect of #572 not being done (yet?) :wink:

aszasz:
May I submit a pull request adding the followng line to README.md (it would be the first line of Setup section)?
**Raspberry Pi 4 Model B:** display must be connected to HDMI port closest to power jack.

JamesH65:
You can use either port, UNLESS you want 4kp60, when you need to use HDMI0.  I think that is the only limitation.

procount:
@JamesH65 - Yes, it's using the legacy driver. Tell me how to mirror the screen to both HDMI ports in Qt and I'll put it in PINN too.

JamesH65:
Mirroring on the legacy driver is possible I think, but not sure currently. A way to do it would be the same same bitmaps for each output, but that would encounter problem if you had two displays of differing resolution. But why mirror, don't you just use the attached display? Where ever it is connected? tvservice -l will list the attached displays and their ID's, so whatever you are using for final output (after QT) can just use that?

procount:
With NOOBS being written in QT, I think QT takes care of all the video output, which is managed by including various QT libraries and setting the output modes with help of the QWS_DISPLAY environment variable. I worked out how to alter this to output to both the default display (usually HDMI0) and VNC simultaneously, but I've not worked out how to output to dual HDMI with Qt (see also #572). 

I suppose for dual displays with the same resolution, it would be possible to use some external fb copy program, like is done for some of the smaller spi type displays, but I think using qt's facilities would provide a better solution - I think it should be possible, I just don't know how to do it.

procount:
[...]

Issue #572, opened 2020-02-10
Enable dual HDMI output


meltwater2000:
In order to keep things simple for beginners it would be ideal if NOOBS was able to output on both HDMI ports.  This would avoid the failure where someone flashes their SD-Card and only sees the Rainbow screen because they used the "wrong" HDMI port (this just makes it look like the SD-Card is broken or firmware issue).  Even pressing 1-4 won't help with this.

Note: The HDMI furthest away will often need to be used if a slimline HDMI adaptor isn't available, so perhaps the default HDMI choice isn't ideal.

If possible dual-screen output would be a better solution, selection via numeric buttons or different choice of default HDMI.

procount:
I'm willing to experiment with this, but where do I start?
```
max_framebuffers=2
+BR2_PACKAGE_QT_GFX_MULTISCREEN=y
Add "Multi:" to QWS_DISPLAY + probably some other strange incantations
API to switch screens in QT?
```

lurch:
It's not quite "dual display" but I guess #606 contains a possible workaround for this.

PR #578, opened 2020-03-25
Document that display must connect to HDMI0 port

aszasz:
While  Dual video is not enabled (github #572) for Pi 4B, warn which HDMI port must be used (github #577)


lurch:
LGTM. @XECDesign ?

XECDesign:
Thank you

PR #584, opened 2020-04-29
Improve INSTRUCTIONS-README.txt

andrum99:
Various improvements, principally:

- Correct minimum SD card size from 8GB to 16GB
- composite video port is not always RCA, so mention it is RCA or 4-pole 3.5mm jack
- Clarify advice on where the files should be on the SD card, i.e. not in a subdirectory.

### DIFF
--- a/sdcontent/INSTRUCTIONS-README.txt
+++ b/sdcontent/INSTRUCTIONS-README.txt
@@ -17,7 +17,7 @@ NOOBS INSTALLATION INSTRUCTIONS
       i. We recommend using gparted (or the command line version, parted)
       ii. Format the entire card as FAT32
    d. Cards over 32GB
-      i. Follow the instructions on https://www.raspberrypi.org/documentation/installation/sdxc_formatting.md
+      i. Follow the instructions at https://www.raspberrypi.org/documentation/installation/sdxc_formatting.md
 3. Extract the files contained in this NOOBS zip file.
 4. Copy the extracted files onto the SD card that you just formatted. The files must be in the root directory of the SD card, not in a subdirectory within this. If the files are in a subdirectory on the SD card, then you will not be able to start a Raspberry Pi from the SD card.
 5. Insert the SD card into the Raspberry Pi and connect the power supply.
--- a/sdcontent/INSTRUCTIONS-README.txt
+++ b/sdcontent/INSTRUCTIONS-README.txt
@@ -28,7 +28,7 @@ Your Pi will now boot into NOOBS and should display a list of operating systems 
 If your display remains blank, you should select the correct output mode for your display by pressing one of the following number keys on your keyboard:
 1. HDMI mode - this is the default display mode.
 2. HDMI safe mode - select this mode if you are using the HDMI connector and cannot see anything on screen when the Pi has booted.
-3. Composite PAL mode - select either this mode or composite NTSC mode if you are using a screen connected to the composite video port (either RCA or 'headphone' connector).
+3. Composite PAL mode - select either this mode or composite NTSC mode if you are using a screen connected to the composite video port. (Either RCA or 'headphone' connector, depending on model of Raspberry Pi).
 4. Composite NTSC mode
 
 If you are still having difficulties after following these instructions, you can ask for help on the Raspberry Pi forums at https://raspberrypi.org/forums.
--- a/sdcontent/INSTRUCTIONS-README.txt
+++ b/sdcontent/INSTRUCTIONS-README.txt
@@ -22,10 +22,11 @@ NOOBS INSTALLATION INSTRUCTIONS
 4. Copy the extracted files onto the SD card that you just formatted. The files must be in the root directory of the SD card, not in a subdirectory within this.
 5. Insert the SD card into the Raspberry Pi and connect the power supply.
 
-**On the Raspberry Pi 4B**:
-
+**Raspberry Pi 4B**:
 - If you are using an HDMI display,  you must connect it to the micro HDMI port closest to the power jack, labelled HDMI 0.
-- If you are instead connecting a composite video display, you must first edit the file config.txt 
+- If you are instead connecting a composite video display, you must first edit the file config.txt
+- You cannot use both an HDMI and composite display at the same time
+
 
 Your Pi will now boot into NOOBS and should display a list of operating systems that you can choose to install.
 If your display remains blank, you should select the correct output mode for your display by pressing one of the following number keys on your keyboard:
--- a/sdcontent/INSTRUCTIONS-README.txt
+++ b/sdcontent/INSTRUCTIONS-README.txt
@@ -11,7 +11,7 @@ NOOBS INSTALLATION INSTRUCTIONS
       i. Download the SD Association's Formatting Tool from https://www.sdcard.org/downloads/formatter_4/eula_mac/
       ii. Install and run the Formatting Tool on your machine
       iii. Select "Overwrite Format"
-      iv. Check that the SD card you inserted matches the one selected by the tool
+      iv. Check that the SD card you inserted matches the one selected by the Formatting Tool
       v. Click the "Format" button
    c. Linux
       i. We recommend using gparted (or the command line version, parted)
--- a/sdcontent/INSTRUCTIONS-README.txt
+++ b/sdcontent/INSTRUCTIONS-README.txt
@@ -32,7 +32,7 @@ Your Pi will now boot into NOOBS and should display a list of operating systems 
 If your display remains blank, you should select the correct output mode for your display by pressing one of the following number keys on your keyboard:
 1. HDMI mode - this is the default display mode.
 2. HDMI safe mode - select this mode if you are using the HDMI connector and cannot see anything on screen when the Pi has booted.
-3. Composite PAL mode - select either this mode or composite NTSC mode if you are using a screen connected to the composite video output. On most Raspberry Pis the composite video output is on the 4-pole 3.5mm jack, along with stereo audio. On the original Raspberry Pi, released in 2012, composite video is output on the RCA connector. Not all 4-pole 3.5mm cables are suitable for using to connect composite video; for details of which cable to use see https://www.raspberrypi.org/documentation/setup/monitor-connection.md .
+3. Composite PAL mode - select either this mode or composite NTSC mode if you are using a screen connected to the composite video output. On most Raspberry Pis the composite video output is on the AV jack, along with stereo audio. On the original Raspberry Pi, released in 2012, composite video is output on the RCA connector. Not all AV cables are suitable for using to connect composite video; for details of which cable to use see https://www.raspberrypi.org/documentation/setup/monitor-connection.md .
 4. Composite NTSC mode
 
 If you are still having difficulties after following these instructions, you can ask for help on the Raspberry Pi forums at https://www.raspberrypi.org/forums .
--- a/sdcontent/INSTRUCTIONS-README.txt
+++ b/sdcontent/INSTRUCTIONS-README.txt
@@ -20,7 +20,7 @@ NOOBS INSTALLATION INSTRUCTIONS
       i. Follow the instructions on https://www.raspberrypi.org/documentation/installation/sdxc_formatting.md
 3. Extract the files contained in this NOOBS zip file.
 4. Copy the extracted files onto the SD card that you just formatted. The files must be in the root directory of the SD card, not in a subdirectory within this. If the files are in a subdirectory on the SD card, then you will not be able to start a Raspberry Pi from the SD card.
-5. Insert the SD card into your Raspberry Pi and connect the power supply.
+5. Insert the SD card into the Raspberry Pi and connect the power supply.
 
 **Note:** on the Raspberry Pi 4B you must connect the display to the micro HDMI port closest to the power jack, labelled HDMI 0.
 
--- a/sdcontent/INSTRUCTIONS-README.txt
+++ b/sdcontent/INSTRUCTIONS-README.txt
@@ -11,7 +11,7 @@ NOOBS INSTALLATION INSTRUCTIONS
       i. Download the SD Association's Formatting Tool from https://www.sdcard.org/downloads/formatter_4/eula_mac/
       ii. Install and run the Formatting Tool on your machine
       iii. Select "Overwrite Format"
-      iv. Check that the SD card you inserted matches the one selected by the Tool
+      iv. Check that the SD card you inserted matches the one selected by the tool
       v. Click the "Format" button
    c. Linux
       i. We recommend using gparted (or the command line version, parted)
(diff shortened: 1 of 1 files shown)


sdcontent/INSTRUCTIONS-README.txt
```diff
@@ -14,21 +14,21 @@ NOOBS INSTALLATION INSTRUCTIONS
...
 Your Pi will now boot into NOOBS and should display a list of operating systems that you can choose to install.
 If your display remains blank, you should select the correct output mode for your display by pressing one of the following number keys on your keyboard:
 1. HDMI mode - this is the default display mode.
 2. HDMI safe mode - select this mode if you are using the HDMI connector and cannot see anything on screen when the Pi has booted.
-3. Composite PAL mode - select either this mode or composite NTSC mode if you are using the composite RCA video connector.
+3. Composite PAL mode - select either this mode or composite NTSC mode if you are using a screen connected to the composite video port. (Either RCA or 'headphone' connector, depending on model of Raspberry Pi).
 4. Composite NTSC mode
 
-If you are still having difficulties after following these instructions, then please visit the Raspberry Pi Forums ( http://www.raspberrypi.org/forums/ ) for support.
+If you are still having difficulties after following these instructions, you can ask for help on the Raspberry Pi forums at https://raspberrypi.org/forums.
```
lurch, line 34: I think the `www` part of the URL ought to be kept, and perhaps it's worth putting a space before the full-stop to reduce the chances of the user trying to go to `https://www.raspberrypi.org/forums.` ?

sdcontent/INSTRUCTIONS-README.txt
```diff
@@ -14,21 +14,21 @@ NOOBS INSTALLATION INSTRUCTIONS
...
 
-**Raspberry Pi 4 Model B:** display must be connected to HDMI port closest to power jack.
+**Note:** on the Raspberry Pi 4B you must connect the display to the micro HDMI port closest to the power jack, labelled HDMI 0.
 
 Your Pi will now boot into NOOBS and should display a list of operating systems that you can choose to install.
 If your display remains blank, you should select the correct output mode for your display by pressing one of the following number keys on your keyboard:
 1. HDMI mode - this is the default display mode.
 2. HDMI safe mode - select this mode if you are using the HDMI connector and cannot see anything on screen when the Pi has booted.
-3. Composite PAL mode - select either this mode or composite NTSC mode if you are using the composite RCA video connector.
+3. Composite PAL mode - select either this mode or composite NTSC mode if you are using a screen connected to the composite video port. (Either RCA or 'headphone' connector, depending on model of Raspberry Pi).
```
lurch, line 31: Do you think it's worth mentioning here about enable_tvout being required to use composite video on the Pi4? WDYT @procount ?

sdcontent/INSTRUCTIONS-README.txt
```diff
@@ -14,21 +14,21 @@ NOOBS INSTALLATION INSTRUCTIONS
...
-      ii. Format the entire disk as FAT32
+      i. We recommend using gparted (or the command line version, parted)
+      ii. Format the entire card as FAT32
    d. Cards over 32GB
-      i. Follow the instructions on https://www.raspberrypi.org/documentation/installation/sdxc_formatting.md
+      i. Follow the instructions at https://www.raspberrypi.org/documentation/installation/sdxc_formatting.md
 3. Extract the files contained in this NOOBS zip file.
-4. Copy the extracted files onto the SD card that you just formatted so that this file is at the root directory of the SD card. Please note that in some cases it may extract the files into a folder, if this is the case then please copy across the files from inside the folder rather than the folder itself.
-5. Insert the SD card into your Pi and connect the power supply.
+4. Copy the extracted files onto the SD card that you just formatted. The files must be in the root directory of the SD card, not in a subdirectory within this. If the files are in a subdirectory on the SD card, then you will not be able to start a Raspberry Pi from the SD card.
```
lurch, line 22: "start a Raspberry Pi" -> "boot a Raspberry Pi" ? :shrug:

lurch: Hopefully this text won't cause users to think they've done something wrong when they see the `os` subdirectory?

sdcontent/INSTRUCTIONS-README.txt
```diff
@@ -1,11 +1,11 @@
 NOOBS INSTALLATION INSTRUCTIONS
 
-1. Insert an SD card that is 8GB or greater in size into your computer.
+1. Insert an SD card that is 16GB or greater in size into your computer.
 2. Format the SD card using the platform-specific instructions below:
    a. Windows (32GB cards and under)
       i. Download the SD Association's Formatting Tool from https://www.sdcard.org/downloads/formatter_4/eula_windows/
       ii. Install and run the Formatting Tool on your machine
-      iii. Check that the SD card you inserted matches the one selected by the Tool
+      iii. Check that the SD card you inserted matches the one selected by the tool
```
lurch, line 8: perhaps "tool" -> "Formatting Tool" for consistency with the previous line? (and then the same change in the Mac instructions)

andrum99: Don't think so. That would make it more wordy. It should be obvious which tool this is referring to, since it was just named in the sentence which immediately precedes it.

sdcontent/INSTRUCTIONS-README.txt
andrum99: Good point. I'm not sure how NOOBS handles that. Does the parameter get passed down to Raspbian when it boots? It would be easier for inexperienced users of they could just add a file to the NOOBS partition to enable composite, like for SSH. Although NOOBS is effectively EOL now, as I understand it.

sdcontent/INSTRUCTIONS-README.txt
andrum99: I've removed the bit about the Pi not booting if there are files in a subdirectory, as I think that was potentially problematic, as you said.

sdcontent/INSTRUCTIONS-README.txt
lurch: > I'm not sure how NOOBS handles that. Does the parameter get passed down to Raspbian when it boots?

I dunno, I can't even remember the last time I used the composite output on any of my Pis, and I've never used composite output on my Pi 4B :confused: 
I pinged @procount because I think this is an area he might have more familiarity with?

> Although NOOBS is effectively EOL now, as I understand it.

Yeah, the way I understand it, the new Raspberry Pi Imager tool is now recommended over NOOBS.

sdcontent/INSTRUCTIONS-README.txt
```diff
@@ -14,21 +14,21 @@ NOOBS INSTALLATION INSTRUCTIONS
       iv. Check that the SD card you inserted matches the one selected by the Tool
```
lurch, line 14: Maybe you want to change "Tool" -> "tool" here, for consistency? :wink:

sdcontent/INSTRUCTIONS-README.txt
procount: Viva PINN! 
Even I recommend using RPi Imager to flash PINN (See https://www.raspberrypi.org/forums/viewtopic.php?f=63&t=142574&start=975#p1638508) as it avoids so many issues with FAT32.
If the aim is just to install ONE OS, the RPIimager can be recommended. But it doesn't replace NOOBS' or PINN's abiity to multi-boot. Horses for courses!

(However, I have had a couple of instances where RPI imager will fail to verify, but Etcher verifies the exact same image 🤷‍♂️)

Re: RCA output - the behaviour has changed with Pi4, so I think it should be mentioned to avoid people wondering why RCA doesn't work without setting enable_tvout. 

Likewise the fact that NOOBS/PINN only works on HDMI-0 (not HDMI-1). It's a common issue on the forums.

andrum99: Thanks for the clarification.

By 'RCA' you of course mean 'TRRS' 😉

procount: Yeah, yeah - sorry - just reusing the above terminology without thinking :)

lurch: > (However, I have had a couple of instances where RPI imager will fail to verify, but Etcher verifies the exact same image man_shrugging)

If that's reproducible with the latest v1.2 of RPi Imager could you please create an issue at https://github.com/raspberrypi/rpi-imager ?

lurch: > Likewise the fact that NOOBS/PINN only works on HDMI-0 (not HDMI-1). It's a common issue on the forums.

Do you want to file a PR against https://www.raspberrypi.org/documentation/installation/noobs.md ? :slightly_smiling_face: 

As a side-note, I guess there's a bit of a balancing act between "how much" gets documented at each of
 * https://github.com/raspberrypi/noobs/blob/master/README.md
 * https://github.com/raspberrypi/noobs/blob/master/sdcontent/INSTRUCTIONS-README.txt
 * https://www.raspberrypi.org/documentation/installation/noobs.md
 * https://www.raspberrypi.org/help/noobs-setup/2/

andrum99: I was taking the view that whatever is in the readme should stay, since the user may not have an internet connection when they do the install

lurch: > whatever is in the readme should stay

Absolutely! I wasn't suggesting removing any documentation, just acknowledging that it'll be a bit of a juggling act deciding how much to duplicate between the various locations that I was pointing at.

sdcontent/INSTRUCTIONS-README.txt
andrum99: Thanks. Incorrect use of proper nouns is one of my bugbears. Grrr.

andrum99:
We have a problem: I've just extracted the NOOBS installation files to have a look to check if config.txt is included, and it is not. How does the user specify enable_tvout=1 on NOOBS? Is it actually possible?

sdcontent/INSTRUCTIONS-README.txt
andrum99: > Do you want to file a PR against https://www.raspberrypi.org/documentation/installation/noobs.md ? 🙂

That page appears to be an older version of https://www.raspberrypi.org/help/noobs-setup/2/.

Also, looking at https://www.raspberrypi.org/help/, it has some very useful content, but it is spread across 3 separate silos:

- raspberrypi.org/help
- raspberrypi.org/learning
- projects.raspberrypi.org

and has a copy of the content from raspberrypi.org/documentation/ at the bottom of it.

So including the documentation section of the website, that's 4 separate information silos. That seems suboptimal to me. Also, each silo has different navigation and style. In particular, the menu across the top of the page is different on each silo.

There appears to be a very small number of pages in the 'help' section of the website. Perhaps it should be merged with the 'documentation' section, using the style of the 'help' section, since that style is more in keeping with the rest of the website, now the blog has been redesigned.

sdcontent/INSTRUCTIONS-README.txt
andrum99: Actually I've changed my mind on that one.

lurch:
> How does the user specify enable_tvout=1 on NOOBS? Is it actually possible?

AFAIK the user would just need to create a new `config.txt` file?

sdcontent/INSTRUCTIONS-README.txt
lurch: > that's 4 separate information silos. That seems suboptimal to me.

Agreed. But AFAIK raspberrypi.org/documentation/ is the only part updated by Raspberry Pi Trading (via the https://github.com/raspberrypi/documentation repo), everything else is managed by Raspberry Pi Foundation.

sdcontent/INSTRUCTIONS-README.txt
```diff
@@ -1,34 +1,38 @@
...
       ii. Install and run the Formatting Tool on your machine
-      iii. Check that the SD card you inserted matches the one selected by the Tool
+      iii. Check that the SD card you inserted matches the one selected by the Formatting Tool
       iv. Click the "Format" button
    b. Mac (32GB cards and under)
       i. Download the SD Association's Formatting Tool from https://www.sdcard.org/downloads/formatter_4/eula_mac/
       ii. Install and run the Formatting Tool on your machine
       iii. Select "Overwrite Format"
-      iv. Check that the SD card you inserted matches the one selected by the Tool
+      iv. Check that the SD card you inserted matches the one selected by the tool
```
lurch, line 14: Sorry to nit-pick, but I think the language here should match line 8 i.e. "tool" -> "Formatting Tool".

sdcontent/INSTRUCTIONS-README.txt
```diff
@@ -1,34 +1,38 @@
...
 
 Your Pi will now boot into NOOBS and should display a list of operating systems that you can choose to install.
 If your display remains blank, you should select the correct output mode for your display by pressing one of the following number keys on your keyboard:
 1. HDMI mode - this is the default display mode.
 2. HDMI safe mode - select this mode if you are using the HDMI connector and cannot see anything on screen when the Pi has booted.
-3. Composite PAL mode - select either this mode or composite NTSC mode if you are using the composite RCA video connector.
+3. Composite PAL mode - select either this mode or composite NTSC mode if you are using a screen connected to the composite video output. On most Raspberry Pis the composite video output is on the 4-pole 3.5mm jack, along with stereo audio. On the original Raspberry Pi, released in 2012, composite video is output on the RCA connector. Not all 4-pole 3.5mm cables are suitable for using to connect composite video; see the [Monitor Connection](https://www.raspberrypi.org/documentation/setup/monitor-connection.md) page for details of which cable to use.
```
lurch, line 35: Probably doesn't make sense to use markdown link-syntax inside a plain-text file? :wink:

andrum99:
> > How does the user specify enable_tvout=1 on NOOBS? Is it actually possible?
> 
> AFAIK the user would just need to create a new `config.txt` file?

OK. I'm going to see if I can find a suitable cable to try this. I've only ever used composite on the original Pi, which uses an RCA jack.

sdcontent/INSTRUCTIONS-README.txt
procount: > > (However, I have had a couple of instances where RPI imager will fail to verify, but Etcher verifies the exact same image man_shrugging)
> 
> If that's reproducible with the latest v1.2 of RPi Imager could you please create an issue at https://github.com/raspberrypi/rpi-imager ?

@lurch - I updated to v1.2 and it seems to be ok now, so either a buggy old version or corrupt installation.  But that's based on only one sample point so far.

lurch: Yeah, IIRC there was a bug in versions of RPi Imager before 1.2 where the verification would (incorrectly) fail if the image-size wasn't an exact-multiple of some blocksize.

andrum99:
OK, so we have 3 options with this one:

- remove references to composite from this file, since it is no longer working. (Or perhaps signpost users to newest version of NOOBS that does work with composite)
- leave as is
- fix NOOBS, but since it is now deprecated / EOL I'm not sure that's worth it.

My preference would be to leave as is, and just change 8GB to 16GB, which was the original motivation for this PR.

Thoughts?

lurch:
> My preference would be to leave as is, and just change 8GB to 16GB, which was the original motivation for this PR.

Sounds like the best option to me :+1:

andrum99:
Closed, as that's the easiest way to do it. See new PR for 8GB -> 16GB.

PR #600, opened 2020-07-29
Update README.md

DaniW42:
Update Wifi Dongle Link

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ If you want to install RISC OS, use NOOBS-Lite instead, or change the included o
 
 #### OS Network Download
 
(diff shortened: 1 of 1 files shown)


procount:
You could also add "or later" after model 3B to account for the Pi4 and possible later models that also have built-in wifi.

lurch:
...or perhaps the USB Wifi Dongle links should be removed, as it's been EOL for a couple of years now...? :man_shrugging:

> You could also add "or later" after model 3B to account for the Pi4 and possible later models that also have built-in wifi.

Don't forget about the Pi Zero W :wink: I think the official RPI documentation uses a suitably generic phrase like "models with built-in wireless".

XECDesign:
Although there hasn't been a response, this still fixes 404s, so I'll merge it.

PR #604, opened 2020-08-26
Fix cert errors

AaronDewes:
I hope it works now, I couldn't test on QT4, @XECDesign can you please do that?


XECDesign:
Seems to build and work. Thank you.

PR #606, opened 2020-11-09
Added auto HDMI switching

procount:
This modification allows NOOBS to automatically switch to the correct HDMI port according to which is connected to a monitor.
The precedence is: DSI, HDMI0, HDMI1.
I'm not sure if line recovery/init#42 is absolutely necessary, but I left it in. 

(I also included setting PS1 in the root/.profile file, which I think helps anyone who decides to use the shell, but delete if you want)


con2fbmap/Makefile
```diff
@@ -0,0 +1,10 @@
+CC=$(PWD)/output/host/usr/bin/arm-buildroot-linux-uclibcgnueabi-gcc
```
lurch, line 1: This doesn't look right - doesn't Buildroot set the correct $CC automatically?

buildroot/package/recovery/init
```diff
@@ -33,6 +33,30 @@ echo 2 >/sys/module/vt/parameters/cur_default
 # Enable syslog
 /etc/init.d/S01logging start > /dev/null
 
+present=`tvservice -l | sed -n -e 's|^Display Number \(.*\),.*$|\1| p'`
+for dev in $present; do
+    fitted=`tvservice -n -v $dev 2>/dev/null`
+    if [[ "$fitted" != "" ]]; then
+        case $dev in
```
lurch, line 40: Given that the values for case 0 and case 2 are the same, perhaps it would make sense to have those set as default outside the loop, and then just switch the case 7 to an `if` instead of a `case` ? :shrug:

buildroot/package/con2fbmap/con2fbmap.mk
```diff
@@ -0,0 +1,25 @@
+#############################################################
+#
+# con2fbmap system
+#
+#############################################################
+
+
+CON2FBMAP_VERSION = 1.0
+CON2FBMAP_SITE = $(TOPDIR)/../con2fbmap
```
lurch, line 9: Probably makes sense to move the conf2bmap files down into this subdirectory, rather than adding an extra directory at the top-level? (but it's been a long time since I've messed around with Buildroot, so accept my apologies if that wouldn't work!)

lurch:
I've not tested this, but if it works I guess it'd be super-useful for Pi400 owners, where the HDMI0 / HDMI1 distinction isn't labelled :slightly_smiling_face:

lurch:
> (I also included setting PS1 in the root/.profile file, which I think helps anyone who decides to use the shell, but delete if you want)

It's up to @XECDesign , but it might be worth making that a separate PR? Personally, I always prefer separate PRs for seperate "things".

For those of us with rusty PS1 knowledge, could you tell us what the shell prompt looks like before and after this change? :shell:

con2fbmap/Makefile
procount: I just used the patch that Floris pointed me at. I didn't modify anything (apart from keeping it separate from the fbset pacakge) and it seemed to work.

buildroot/package/recovery/init
procount: Maybe that would work too. I wanted to keep it open in case any of the other values were needed too. I did think of reversing the loop nesting to allow different priorities, but I decided the default were sufficient.

buildroot/package/con2fbmap/con2fbmap.mk
procount: I put it at the same level as my rebootp package in PINN, keeping it separate from buildroot.

procount:
> I've not tested this, but if it works I guess it'd be super-useful for Pi400 owners, where the HDMI0 / HDMI1 distinction isn't labelled

That's where the idea came from. I built and tested it in NOOBS with DSI touchscreen, HDMI0 and HDMI1 connection combinations. It's already in PINN and working nicely. I haven't tested it with my pinn_init script for configuring the pimoroni hyperpixels, but those drivers need looking at anyway.

>  it might be worth making that a separate PR?

I thought about that, and I remembered that thought again, just after I'd squashed the commits ;) But hey, I can separate it out again if necessary.

Before this change the shell prompt is just a '#' symbol. Nothing else.
After this change, it looks more like the standard `User@hostname: <pwd> #` . It makes it a lot easier to know where you are in the file tree.

con2fbmap/Makefile
maxnet: Eh, didn't I point you to a patch that Debian uses to add fbcon to the Debian package?!
I think it is highly unlikely Debian sets a compiler to arm-buildroot :-)

buildroot/package/con2fbmap/con2fbmap.mk
```diff
@@ -0,0 +1,25 @@
...
+CON2FBMAP_SITE_METHOD = local
+CON2FBMAP_LICENSE = BSD-3c
+CON2FBMAP_LICENSE_FILES = LICENSE.txt
```
maxnet, line 12: You cherry picked some files from another project, but forget to copy the file that describes the license the files are under.

And fbset looks more like GPL-2.0 rather than BSD-3c

buildroot/package/recovery/init
lurch: Fair enough, no harm in keeping it easy-to-extend.

buildroot/package/con2fbmap/con2fbmap.mk
lurch: Fair enough, I'll defer this decision to @maxnet / @XECDesign

maxnet: >Fair enough, I'll defer this decision to @maxnet / @XECDesign

Not mine to make.
Just dropping by here as normal bystander, as someone mentioned my name :-)
Have not been involved with NOOBS in recent years.

lurch: > Just dropping by here as normal bystander, as someone mentioned my name :-)

:slightly_smiling_face: :wave: 

> Have not been involved with NOOBS in recent years.

Ditto. I was only mentioning your name because you have more recent buildroot experience than me. Apologies if I gave the wrong impression.

con2fbmap/Makefile
procount: Sorry, I obviously got confused. Mea maxima culpa! Apologies.
I'll take it out and rebuild just to make sure.

buildroot/package/con2fbmap/con2fbmap.mk
procount: Thanks. I'm not very proficient with licenses. I'll update appropriately.

lurch: Unfortunately it's something you need to be quite careful with, because some people get very upset if you get it wrong.

con2fbmap/Makefile
procount: If I take this line out, it just uses `cc` and then `strip` fails to recognise the format of the input file (cos I think it would have built it for x86).
So I don't know why it's not picking up the buildroot macros. Any ideas?

maxnet: In the buildroot package file try if it works better with just:

```
define CON2FBMAP_BUILD_CMDS
	$(MAKE) $(TARGET_CONFIGURE_OPTS) -C $(@D)
endef
```

Might also want to simplify the Makefile to say:

```
con2fbmap: con2fbmap.c
	$(CC) con2fbmap.c -o con2fbmap
```

- First target in Makefile is default target. Can leave out the "all" stuff.
- Don't think it is necessary to compile statically.

procount: Something else must be wrong, maybe in con2fbmap.mk?
1, If I leave out the all: target, it complains and won't build at all.
2. Even with the above changes, it still just uses `cc`, so fails.

maxnet: 1) Make sure you do the change to the buildroot .mk first.
You had "$(MAKE) -C $(@D) all" first, and that does mention "all" explicitly.

If you had "$(MAKE) -C $(@D)" without "all", it would use default target.

That will still not set CC and LD for you though.
That is what the $(TARGET_CONFIGURE_OPTS) does set, so hence my recommended change:

```   
$(MAKE) $(TARGET_CONFIGURE_OPTS) -C $(@D)
```

That is just one way to Rome though.
Could also use:

```
$(MAKE) CC="$(TARGET_CC)" LD="$(TARGET_LD)" -C $(@D)
```

Would give the same result.
TARGET_CONFIGURE_OPTS sets CC and LD and a bunch options more.
See for list: https://github.com/raspberrypi/noobs/blob/18db2628678e54d1a431fe6f21dfb9e19a962425/buildroot/package/Makefile.in#L249

procount: Doh! OK, I see where I went wrong. Many thanks for your patience.

buildroot/package/recovery/recovery.mk
```diff
@@ -33,6 +33,7 @@ define RECOVERY_INSTALL_TARGET_CMDS
 	$(INSTALL) -m 0755 package/recovery/unicode-fonts/NanumBarunGothicBold.ttf $(TARGET_DIR)/usr/lib/fonts/NanumBarunGothicBold.ttf
 	$(INSTALL) -m 0755 package/recovery/data/data $(TARGET_DIR)/usr/data
 	$(INSTALL) -m 0644 $(@D)/cmdline.txt $(BINARIES_DIR)/cmdline.txt
+
```
lurch, line 36: Could remove this change?

procount: yeah, small remnant from separating the profile PR.

buildroot/package/rpi-firmware/config.txt
```diff
@@ -5,3 +5,4 @@ fixup_file=fixup_rc.dat
 [pi4]
 start_file=recover4.elf
 fixup_file=fixup4rc.dat
+max_framebuffers=2
```
lurch, line 8: I googled and found https://www.raspberrypi.org/forums/viewtopic.php?t=245789 which says that `max_framebuffers` has to match the number of physical displays. So does this 'break' things when only a single display is attached? :shrug:

procount: That answer of "matching" was not confirmed by JamesH. 
AIUI, it is the **maximum** number of framebuffers that could be allocated, so it can be greater than or equal to the number of displays, although setting it greater might be  a waste of resources.
Certainly if it is less than the number of displays, you won't get the framebuffer.
You might be able to use HDMI1 with only 1 framebuffer if you set framebuffer_priority, but I've not tried it, but that would also mean the user needs to set it in config.txt, rather than NOOBS dynamically using which ever is connected. So you need to set max_framebuffers=2 to access HDMI1 with this technique.

lurch: > it can be greater than or equal to the number of displays, although setting it greater might be a waste of resources.

Ah cool. Obviously "wasting resources" could be an issue on Raspberry Pi OS, but isn't an issue for a limited environment like NOOBS. Sounds like you've already tested it thoroughly :+1: 
I hope you didn't mind me asking!

procount: I always welcome your pedantic mindset! 😄
It's already been released in PINN and seems to work quite well. I think it had the most downloads immediately after a release, so users must like the feature.
But all the comments here have enabled me to improve things further (especially the builds) and I've learnt some more stuff along the way. So thanks to all.

lurch: FYI: https://github.com/raspberrypi/documentation/commit/4f691ff4cc96eb938631caf5924005eb8817a401

procount: Like I said.

buildroot/package/con2fbmap/con2fbmap.mk
ret394: > 
> 
> Probably makes sense to move the con2fbmap files down into this subdirectory, rather than adding an extra directory at the top-level? (but it's been a long time since I've messed around with Buildroot, so please accept my apologies if that wouldn't work!)

Same for any weird reviews from me. I'll keep non tech comments separate to make it easy. 

But those non tech ones may end up to be tech comments if you think about it.

ret394:
> where the HDMI0 / HDMI1 distinction isn't labelled

Why?🤦‍♀️

buildroot/package/rpi-firmware/config.txt
ret394: u guys should read the datasheets first, I know they are primitive. 

But you could ask for help. I feel James got the production ones 😏

PS - James, transfer it to a copy with the notes. Don't tell them about the piece of scrap. They cannot Google it.

ret394:
James, why no `=3`?
 https://github.com/raspberrypi/documentation/commit/4f691ff4cc96eb938631caf5924005eb8817a401

XECDesign:
Many thanks!

Issue #608, opened 2020-11-16
Recovery Wallpaper


duckida:
Hi. I checked the wallpaper([/recovery/wallpaper/png](https://github.com/duckida/noobs/blob/master/recovery/wallpaper.png)) and noticed it was **from 2013!!!**. I have created a more modern one:
![New Wallpaper](https://raw.githubusercontent.com/duckida/TestWallpp/main/Screen%20Shot%202020-11-16%20at%2012.58.27%20PM.png)

Can we please use this new one? It will be the first step in making NOOBS consistent with Raspberry Pi OS!

duckida:
@lurch Please See

duckida:
Feel free to make your own recovery wallpapers and submit them

lurch:
@duckida It's @XECDesign 's decision to make, not mine :wink:

XECDesign:
These kinds of changes would need to go through the (very busy) design team. We are currently only making basic maintenance changes to NOOBS to keep it working with newer models, so I don't think the wallpaper is really an issue that's likely to get looked at.

duckida:
@XECDesign Please See #610

PR #609, opened 2020-11-16
Change Recovery Splash Screen

duckida:
I removed the old wallpaper and replaced it with my new one to solve #608!


lurch:
`recovery/Screen Shot 2020-11-16 at 12.58.27 PM.png` ??? I think something might have gone wrong with your PR :laughing:

duckida:
> `recovery/Screen Shot 2020-11-16 at 12.58.27 PM.png` ??? I think something might have gone wrong with your PR 😆

Fixed it, @lurch

lurch:
Your new wallpaper is 1026x688, which means it probably wouldn't look very good when viewed at 640x480 ? (don't forget that NOOBS also supports composite-video output)

lurch:
Looks like this branch is no longer relevant to the original PR. Should it be closed @duckida ?

PR #636, opened 2021-10-25
Bump libatomic_ops to 7.6.12

ivmai:
Contains, in particular:
* fix in asm code for x86
* use gcc intrinsic instead of asm code where possible
* eliminate compiler warnings
* support of asan/msan/tsan

### DIFF
--- a/buildroot/package/libatomic_ops/libatomic_ops.mk
+++ b/buildroot/package/libatomic_ops/libatomic_ops.mk
@@ -4,7 +4,7 @@
 #
 ################################################################################
 
-LIBATOMIC_OPS_VERSION = libatomic_ops-7_4_0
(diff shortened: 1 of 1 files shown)


ivmai:
> Some kind of explanation / requirement for this change might be nice?

I've updated the PR description.

lurch:
Hmmm, I've just checked and NOOBS doesn't actually _use_ libatomic_ops (https://github.com/raspberrypi/noobs/blob/master/buildroot/.config#L1368 says `# BR2_PACKAGE_LIBATOMIC_OPS is not set`), so this PR is probably redundant?

ivmai:
Got it. As I understand you periodically sync it with buildroot upstream (https://github.com/buildroot/buildroot).
Closing the PR.

PR #637, opened 2021-10-26
Remove bdwgc duplicate package and update it to 8.0.6

ivmai:
libgc is a synonym of bdwgc, this PR removes libgc package (which is, by the way, not used by any other package)
bdwgc is updated to version 8.0.6.

Major changes:
* Drop upstreamed patches (configure.ac, include/private/gcconfig.h)
* Support aarch64 and m68k architectures
* Delete --with-libatomic-ops=yes (let configure decide about atomic ops)
* Upstream changes: fix CVE-2016-9427 and other bugs


ivmai:
@lurch, this PR seems to be not needed neither, right?

XECDesign:
Yup, I don't think this makes sense for NOOBS.

PR #647, opened 2023-02-26
README.md: Add Link to latest version of NOOBs-Lite

Franz-Heinz:
Adding the link to the latest version of NOOBs-Lite in order to make it more easy to find for users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@ NOOBS is designed to make it easy to select and install operating systems for th
 
 The latest official release of NOOBS can be downloaded from https://downloads.raspberrypi.org/NOOBS_latest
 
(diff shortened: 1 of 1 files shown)


XECDesign:
@ghollingworth should we add a link to PINN and archive the repo?

ghollingworth:
Yes, happy with that.

XECDesign:
@procount would you prefer a link to github or sourceforge?

procount:
Hmm, tricky, but overall I think if you are only posting one link the github page would be better, because there is a link there to download pinn-lite.zip from Sourceforge, but not the other way round.

Or you could put both links up - Github for documentation and source,, and Sourceforge for downloading the zip/img files?

